PR #342, opened 2015-02-10
[SERVDEV-49] - Changing project revision to "INNOVATION-SNAPSHOT"

brosander:
[BACKLOG-435] - Addition of the "angular-translate" angular module
* The module is patched. It had a few bugs that were fixed by us.
   Also added a new feature for being able to use a client-side format-adapter, for translating resource properties files to JSON.
   Changes are here: https://github.com/dcleao/angular-translate/
   Will send pull-requests to the author, but in the mean time...
* common-ui's existing dojo-based messages use a service, "i18n", to translate properties files to JSON.
   This is provided by biserver and is therefore not directly usable in an OSGI bundle in spoon.
   As soon as we find a way to have this service, or a similar one, available for OSGI consumption, we can drop the client-side translation.
* Fix error in JS.
* Fix wrong file names.
* Adding Array.isArray shim for IE8.


brosander:
Moving work Duarte did on innovation branch over to future-develop

PR #343, opened 2015-02-16
[ANALYZER-2933] Fixes tooltip not showing "measure interpolated" suffix for interpolated measures.

SergeyTravin:
@e-cuellar, @mdamour1976  could you merge it?

### DIFF
--- a/package-res/resources/web/vizapi/ccc/ccc_wrapper.js
+++ b/package-res/resources/web/vizapi/ccc/ccc_wrapper.js
@@ -1388,8 +1388,15 @@ function(def, pvc, pv){
                     }
 
                     var suffix;
-                    if(complex.isInterpolated){
-                        suffix = this.chart._message('chartTooltipGemInterp_' + complex.interpolation);
+                    
+                    // It can happen that the scene has more than one datum.
+                    // One is a null one and the other an interpolated one.
+                    // We may receive the null one in `complex` and 
+                    // miss detecting that the scene is actually interpolated.
+                    if(context && context.scene && context.scene.datums()) {
+                    	if( context.scene.datums().where(function(d) { return d.isInterpolated && d.interpDimName === cccDimName; }).first() ) {
+                    		suffix = this.chart._message('chartTooltipGemInterp_' + complexInterp.interpolation);
+                    	}
                     } else if(complex.isTrend/* && atom.label*/){
                         // TODO: "atom.label" -- is a weak test for trended measures,
                         // that relies on the fact that non-trended measures are left null


mdamour1976:
@SergeyTravin  Since this code is in common-ui, the affected areas may be greater than just Analyzer, has it been considered to protected against regressions introduced by possible NPE:
                    var complexInterp = context.scene.datums()
I see the assumption here may be fine if we always have context.scene.datums, but a tiny bit of protection will make me feel slightly more comfortable (especially not knowing this code).

SergeyTravin:
@mdamour1976, Great thanks for important notes. I changed the code. Could you review it again?

mdamour1976:
I was thinking more about each part of this, rather than the whole..
context.scene.datums()

Not that it is possible (I don't know), but if context is null, we can still NPE on context.scene.

And same for context.scene.datums() if context.scene is null.

So we should add some:

if ( context && context.scene ) {
  // do the right thing
}

SergeyTravin:
@mdamour1976, I don't believe that it is possible, but I changed code to avoid this situation at future.

dcleao:
Yep, I can confirm that there is no possibility for a NPE to occur. Of course, if something goes really bad, somewhere upstack, who knows.

SergeyTravin:
@mdamour1976, any comments?

SergeyTravin:
@nbaker, @rfellows, @mdamour1976 Could you provide any feedback?

SergeyTravin:
@dkincade,  @nbaker, @rfellows, @mdamour1976 Could you provide any feedback?

dcleao:
@mdamour1976, @SergeyTravin: upon refactor of code in future develop branch I passed through this commit's code and spotted an error. Sergey's change to avoid NPE introduced a ReferenceError...

The original code was (from https://github.com/dcleao/pentaho-platform-plugin-common-ui/commit/668b473b1e15b6b9eef87538a545b8fb10d93bf7):

```javascript
var complexInterp = context.scene.datums()
    .where (function(d) { return d.isInterpolated && d.interpDimName === cccDimName; })
    .first();
    if(complexInterp) {
        suffix = this.chart._message('chartTooltipGemInterp_' + complexInterp.interpolation);
```

In the new code, variable `complexInterp` is not defined...

To protect against NPE, you can instead do:
```javascript
var complexInterp = context && context.scene && context.scene.datums()
    .where (function(d) { return d.isInterpolated && d.interpDimName === cccDimName; })
    .first();
    if(complexInterp) {
        suffix = this.chart._message('chartTooltipGemInterp_' + complexInterp.interpolation);
```

This issue is planned for 5.4...

PR #345, opened 2015-02-18
[SP-1688] - Backport of PRD-5312 - A number entered on a string text box...

akhayrutdinov:
... parameter is formatted as #,### after tabbing out of the parameter (5.3 Suite) - (cherry picked from commit 04a1e99)

@tmorgner review it please. Backported for 5.3 March SP

### DIFF
--- a/package-res/resources/web/prompting/pentaho-prompting-components.js
+++ b/package-res/resources/web/prompting/pentaho-prompting-components.js
@@ -408,10 +408,14 @@ define("common-ui/prompting/pentaho-prompting-components", ['common-ui/prompting
             initialValue = this.formatter ? this.formatter.format(this.transportFormatter.parse(v.label)) : v.label;
 
             try {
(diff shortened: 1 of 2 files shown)


tmorgner:
Lets wait to hear from Pedro Vale about the JS test-case failures before merging this.

akhayrutdinov:
@tmorgner,
I've cut the promptingSpec.js from the PR. Now the rest of tests wouldn't be impacted.

Since SP sprint is almost over, we should ensure all tasks are completed. The fix works for master (I've already resolved the ticket), thus I believe it would also pan out for 5.3. I'll continue dealing with Pedro to recover JS tests in master.

What do you think of it? Is it an affordable solution?

tmorgner:
Skipping the tests in 5.3 for now should not pose a huge risk as long as we have the tests running in master in the near future. 

@dkincade, @mdamour1976 please merge

PR #361, opened 2015-04-21
[BACKLOG-2467] - Implement a standard method for registering Javascript plugin objects

dcleao:
* Initial **basic** implementation adapted from Nick's prototype.
* Adding common Require-JS properties to the `requireCfg` object
   * Also includes an `requireCfg.config.service` object, for easier configuration of the proposed AMD plugin, whose AMD module id is `service`.
   * Here, just changing `subfloor-js.xml` and `config/context.js` (used for testing); but there's a corresponding commit in "pentaho-platform/extensions", to affect the generation of `webcontext.js`.
   * I know changing "subfloor-js.xml" is always temporary - this commit would serve to propose the change to ENGOPS.

@pentaho-nbaker please review.

### DIFF
--- a/build-res/subfloor-js.xml
+++ b/build-res/subfloor-js.xml
@@ -139,7 +139,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
           </fileset>
         </path>
         <echo message="Collecting RequireJS Configurations"/>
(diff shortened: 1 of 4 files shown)


dcleao:
Associated pentaho platform pull-request: https://github.com/pentaho/pentaho-platform/pull/2281.

dcleao:
This is an initial implementation solely to support VizAPI developments. When we attack this issue, properly, on a sprint, we'll add unit-tests.

PR #371, opened 2015-05-04
[BACKLOG-2540] - AMD Conversion - Adding ES6 Promise Shim

dcleao:
* Support the global ES6 Promise class
* Using for VizAPI's _async_ visualization factories

@pentaho-nbaker:

I know _dojo_ has a Promise/A+ implementation, and a deferred class. However, [ES6 promises](http://www.html5rocks.com/en/tutorials/es6/promises/), being compatible on the consumer side, by conforming to Promises/A+:
* are globally available: `Promise`
* are/will-be standard
* are already implemented in "modern" browsers
* have some helpful static methods as well, like `Promise.all` and others ([Promise on MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise))
* have a different producer contract, that relies solely on the `Promise` constructor (no deferred class), based on [RSVP.js's contract](https://github.com/tildeio/rsvp.js):
```javascript
new Promise(function(resolve, reject) {
    // do work and resolve or reject, sync or async
});
```

So an ES6 promise polyfill/shim seems to be what is desirable now, to bridge the gap into the future.
The library that I'm including is [es6-promise](https://github.com/jakearchibald/es6-promise/), a stripped-down version of `RSVP.js` that contains solely the standard behavior.
Then, I added an AMD module ("es6-promise-shim") that one requires to perform the polyfill/shim. Conveniently, the AMD module returns the Promise class - but one can then assume it is a global, native thing.

This still has to go through the "component approval process", but given that the license is MIT, I'm assuming that this is more of a technical question.

What do you think?

CC: @pminutillo, @DFieldFL, @bennychow, @pamval


dcleao:
Ping @pentaho-nbaker.

pentaho-nbaker:
Totally man. I support not dragging dojo in simply to provide this.

PR #384, opened 2015-06-12
[ANALYZER-3058] - Memory leaks during multiple changing of chart type

dcleao:
* Undoing some of the cleared references and disposed objects as drag and drop operations that destroyed the panel in the drop (Analyzer refreshReport called) would fail
* Not all memory is reclaimed now, but still much better than the initial state.

@diogofscmariano, @pamval, @bennychow please review.


dcleao:
Sorry guys. Merging this due to the urgency of it. However, please make your comments when you can.

PR #405, opened 2015-06-26
[CLEANUP] - Making the require config work with our require osgi bundle

brosander:


brosander:
@dcleao @krivera-pentaho please review, requireCfg.map is undefined when common-ui is used from our osgi require bundle

dcleao:
Hi Bryan. We're upgrading the version of requirejs (http://jira.pentaho.com/browse/COMP-1460) and also, added more variables to the `requireCfg` object, by default (http://jira.pentaho.com/browse/ENGOPS-1097).
You can see the new `requireCfg` structure, here: https://github.com/dcleao/pentaho-platform-plugin-common-ui/commit/15b572576c93c2cb55beae3313bc4bb3831cb8f3#diff-2f0c1c4205fd7c558f3d00b8b0a01d02R18

I think that the correct thing to do would be to change OSGI to have the same structure.

krivera-pentaho:
I have to defer to Duarte's expertis here.

brosander:
https://github.com/pentaho/pentaho-osgi-bundles/pull/41

PR #410, opened 2015-07-02
BACKLOG-4054 & BACKLOG-4061

diogofscmariano:
[BACKLOG-4054]
 - CDF version updated, where references to global CDF global variables were removed.
[BACKLOG-4061]
- Initial commit with first proposal of AMD modules for the prompting engine refactored
  - Builders broken down in different files, one file per builder with dependencies between them
  - Components broken down in different files, one file per component with dependencies between them
- Removed unneeded files
- Unit tests not working due to dependency from CDF not compatible with Common-ui expected directories
- Removed bind shim, no longer needed due to the drop of IE8 support for 6.0
- Tested with debug=true due to build not working at the moment and using just Report Viewer
- Nothing changed in the prompting workflow


diogofscmariano:
@plagoa @annapopova1 @krivera-pentaho @rfellows @scottyaslan @pamval please review

annapopova1:
Looks good for first initial commit.

PR #412, opened 2015-07-03
[BACKLOG-4055]

diogofscmariano:
- Updated the way cdf code is referenced within code stage.
- In cdf the ability to fully configure the location of the cdf source code was added. This configuration was added in context.js
- Fixed unit test for StaticAutocompleteBoxComponent due to missing jquery.ui dependency


diogofscmariano:
@pamval please review

PR #422, opened 2015-07-10
[BACKLOG-4132] - Develop unit tests for builders

plagoa:
- Included unit tests for prompting builders;
- These tests will only test correct building of components, not their behaviour;
- Changed two of the builders to return correct widget (DateInputBuilder and ExternalInputBuilder);


plagoa:
@pamval @annapopova1 @diogofscmariano @rfellows @krivera-pentaho @scottyaslan please review

PR #430, opened 2015-07-13
[BACKLOG-4139]

diogofscmariano:
- Added documentation for Parameter package
- Added license headers
[BACKLOG-4138]
- Removed unneeded method from Parameter
- Removed unneeded dependency Base from files


diogofscmariano:
@pamval @krivera-pentaho @scottyaslan @rfellows @annapopova1 @plagoa

PR #444, opened 2015-07-17
[BACKLOG-4063]

diogofscmariano:
- Added target for js documentation using jsdocs. This target allows multiple file configuration per documented module
- Changed current doc-js to call the new jsdoc target along with the previous one with yuidocs


diogofscmariano:
@pamval @scottyaslan @krivera-pentaho @rfellows @plagoa @annapopova1 please review

PR #446, opened 2015-07-24
[CLEANUP] Karma launcher. Too many test files starts giving file not found errors

dcleao:
* Excluded some more unused files from the karma configs

@plagoa please review


dcleao:
@diogofscmariano please review.

PR #453, opened 2015-08-04
[BACKLOG-4045] - Review Component - Text Field

plagoa:
- Included Dev documentation in the TextInputBuilder

### DIFF
--- a/package-res/resources/web/prompting/builders/TextInputBuilder.js
+++ b/package-res/resources/web/prompting/builders/TextInputBuilder.js
@@ -15,16 +15,57 @@
  *
  */
 
+/**
+ * <h2>The Text Input Builder</h2>
+ *
+ * To use the TextInputBuilder you should require the appropriate file
+ * from Common-Ui:
+ *
+ * <pre><code>
+ *   require(['common-ui/builders/TextInputBuilder'], 
+ *     function(TextInputBuilder) { 
+ *       
+ *     }
+ *   );
+ * </code></pre>
+ * 
+ * To get the component you'll have to create a new instance of the builder and
+ * call the <code>build</code> method:
+ *
+ * <pre><code>
+ *   var textInputBuilder = new TextInputBuilder();
+ *
+ *   var textInputComponent = textInputBuilder.build(args);
+ * </code></pre>
+ *
+ * where 'args' is an object that contains the prompt panel and the parameters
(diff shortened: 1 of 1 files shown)


plagoa:
@diogofscmariano @annapopova1 @rfellows @krivera-pentaho @scottyaslan @pamval please review

diogofscmariano:
@plagoa the link points to the wrong documentation sample

plagoa:
Corrected the documentation sample link. @diogofscmariano please review

package-res/resources/web/prompting/builders/TextInputBuilder.js
```diff
@@ -15,16 +15,53 @@
...
+ * call the <code>build</code> method:
+ *
+ * <pre><code>
+ *   var textInputBuilder = new TextInputBuilder();
+ *
+ *   var textInputComponent = textInputBuilder.build(args);
+ * </code></pre>
+ *
+ * where 'args' is an object that contains the prompt panel and the parameters
+ * necessary for the component as per [the CDF documentation]{@link http://localhost:8080/pentaho/api/repos/:public:plugin-samples:pentaho-cdf:pentaho-cdf-require:30-documentation:30-component_reference:10-core:37-TextInputComponent:text_input_component.xcdf/generatedContent}.
```
rfellows: Do we not have place in the internet where our docs get published that we can link to rather than localhost? Chances are when someone sees the link and tries to follow it, they won't have the server running and or will not have the cdf doc plugin loaded in the platform.

diogofscmariano: Agree with the question. CDF documentation isnt located on any place, so the link can end up not opening. The case of CDF doc not being there is a decision of removing that manually, since they are included by default. Although, I agree with the potential issue. 
Anyway, i think it is best to link with what we have than copy what we already have on CDF.

rfellows: ok.

diogofscmariano: @pamval do you think this is an acceptable commitment?

diogofscmariano: here is another proposal i worked on: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/454
I was thinking... and for some cases it could probably makes sense to have an js object with the example of the args. We could end having that, as an example and also point to the CDF documentation in case someone needs to have more information about extra configuration and/or the types of each property.
Please let me know what you guys think

pamval: agree with the link as long as you call out that the links are for in-product documentation and that they should have the ba server running on the same machine

plagoa:
Included reference to CDF documentation location.

PR #463, opened 2015-08-07
Updated CDF Dashboards update lifecycle to only update necessary components

krivera-pentaho:


krivera-pentaho:
@scottyaslan @diogofscmariano @plagoa @rfellows @tmorgner please review.
I still need to write the unit tests for the new update method, but I fixed all of the failing unit tests due to the changes to the prompt panel before merging with pentaho/master

annapopova1:
Notices about "init" and "update" functions.
1. Now init runs once and still has special logic for working with autoTopValue, autoFocus and scrolling of components.
But new function "update" has not this logic and I afraid we can lose this functional.
2. In the old implementation we had one main funciton "init". This function can be overriden by plugins. For example, PIR change init logic (show/hide hint for prompt panel after init):
```
	this.promptPanel.init = lang.hitch(this, function() {
        panelInit.call(this.promptPanel);
        // Update our UI after Dashboards has rendered all components
        setTimeout(function() {
          if (view.hasVisibleParameters(model.report)) {
            domClass.add("parameterPanelHint", "hidden");
          } else {
            domClass.remove("parameterPanelHint", "hidden");
          }
          view.resize();
        }, /*Dashboards.renderDelay ||*/ 300);
      });
```
But now this functional does not work correctly because init is executed once. As a result we see prompt components in panel and also default hint.
To fix it probably we should fix PIR or work with only "init" and use "update" as a private function or implement something post init logic that can be used by different plugins.

annapopova1:
To not forget also I past comment from @tmorgner about errors:
"For the long run, I have a feeling the error label components can be updated more efficiently than by doing a add/remove cycle.

However, right now the labels are only added on demand - no error on init() and you get no error label. A better approach might be to have them added to all components all the time, and then use CSS visible to show/hide them as needed. That would mean there would be one more DOM element that would stay in place, reducing churn and flickering."

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -569,54 +660,181 @@ define(['amd!cdf/lib/underscore', 'cdf/lib/Base', 'cdf/Logger', 'dojo/number', '
...
+            var params = changeWrap.params;
+
+            for (var i in params) {
+              var param = params[i];
+
+              // Find selected value in param values list and set it. This works, even if the data in valuesArray is different
+              var selectedValues = param.getSelectedValuesValue();
+              for (var j in selectedValues) {
+                this.setParameterValue(param, selectedValues[j]);
+              }
```
diogofscmariano: Is it possible to have more than one selected value? I image that yes. Then, we need to set the list of values and not do this for the list, because in this case, only the last selected value will be the parameter value, which is wrong

pamval:
I think this should be squashed into less commits. Also, a couple of Anna's commits do not follow the standard regarding description so that needs to be changed.

package-res/resources/web/prompting/PromptPanel.js
diogofscmariano: There is another possible optimization here that will completely avoid the component update. 
If we knew that the parameter has the same values, we can compare the parameter that CDF Dashboard has from the previous iteration (because when the component gets the value changed, that change is stored) and if it is the same that we get from the parameters call, then there is no need to even call the update on the dashboard for the component. In case this isnt clear...

1 - user changes the value in a dropdown
2 - parameter change is stored in cdf
3 - a new parameters call is triggered and handled
4 - we found that the parameter is the same and it has one/more values selected (if not, do what we have here now)
5 - the value(s) selected is the same from cdf, dont do anything
6 - the value(s) selected is different, then we set the values and update the component, so its vizual representation gets updated

diogofscmariano:
Please do not forget that the Garbage collector component was removed. please add that jira in the commits

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -27,8 +27,9 @@
  * @property {Dashboard} dashboard The dashboard object assigned to the prompt
  * @property {Boolean} parametersChanged True if the parameters have changed, False otherwise
  */
-define(['amd!cdf/lib/underscore', 'cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/util/util', 'common-ui/util/GUIDHelper', './WidgetBuilder', 'cdf/Dashboard.Clean'],
-    function (_, Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashboard) {
+define(['amd!cdf/lib/underscore', 'cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/util/util', 'common-ui/util/GUIDHelper', './WidgetBuilder', 'cdf/Dashboard.Clean', './parameters/ParameterDefinitionDiffer'],
```
diogofscmariano: This needs to use the underscore lib from common-ui. please follow this conversation https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/467

package-res/resources/web/prompting/PromptPanel.js
krivera-pentaho: Yes, you can have multiple values selected with the multi select list boxes.

krivera-pentaho:
@pamval I could squash it into one commit, but then we would lose track of the 3 different issues. Please advise.

pamval:
One commit per issue ? If that's a lot of work, then nevermind. But the commit messages for those two commits by Anna need to be fixed.

annapopova1:
IMO we can squash all commits as a BACKLOG-4070 except last Kris's commit about GC. Because most of all relates life cycle task.

krivera-pentaho:
ok, i will squash them down.

krivera-pentaho:
Alright, history has been rewritten and the commits have been squashed.

PR #465, opened 2015-08-11
[BACKLOG-4177] - PIR - Filter Specify from the list doesn't work

plagoa:
- changed plugin attributes to use the overriding class loader.


plagoa:
@diogofscmariano @annapopova1 @krivera-pentaho @scottyaslan @rfellows @pamval please review

PR #467, opened 2015-08-11
[BACKLOG-4602] inlcude missing underscore lib

scottyaslan:


diogofscmariano:
I dont like it, but i think this is the right thing to do now. IMO, there is no reason why common-ui needs to use cdf libs, but for now, this is the only way we have to make sure the lib is loaded correctly. @pamval please check

pamval:
common-ui ships its own underscore (https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/underscore/underscore.js). Why not request that ?

diogofscmariano:
@pamval right... we are just doing that for jquery plugins loaded from cdf. in this case we need to load the underscore, although this is strange, because something must be loading underscore, because on my tests this was available in the global scope and i didnt got any error here

pamval:
uff, instead of define("common-ui/util/util", ["amd!cdf/lib/underscore", why not define("common-ui/util/util", ["common-ui/underscore/underscore", ?

Does that not work ?

diogofscmariano:
Actually, makes more sense to be define("common-ui/util/util", ["common-ui/underscore", like the require cfg configs: https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/common-ui-require-js-cfg.js#L105

scottyaslan:
Updated to be define("common-ui/util/util", ["common-ui/underscore"

PR #476, opened 2015-08-15
[BACKLOG-4030]

diogofscmariano:
- Added changes to date picker to allow Dashboards plugin usage
- Added conversion of jquery date format to dojo. Unit tests added to it
- Fixed issue when parameter from dashboards is an array (the use case from Dashboards includes it when the default value is defined in the gwt window of the prompt creation)
- Added formatter from dojo to properly format dates in Dashboards plugin
- Added ability to define start and end date of the date picker (needed for Dashboards plugin)
- Styles still need to be adjusted for Dashboards plugin

### DIFF
--- a/package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
+++ b/package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
@@ -54,10 +54,12 @@
  * @class
  * @extends BaseComponent
  */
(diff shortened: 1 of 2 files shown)


diogofscmariano:
@pamval @scottyaslan @rfellows @annapopova1 @krivera-pentaho @plagoa  please review

PR #478, opened 2015-08-18
[BACKLOG-4030] - Review Component - Date Picker

plagoa:
- Allow default null or empty specified value;
- Allow change of date formats in the date component;
- Some code refactoring;

### DIFF
--- a/package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
+++ b/package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
@@ -89,16 +89,25 @@ define(['cdf/components/BaseComponent', "dojo/date/locale", 'dijit/form/DateText
           var parameterValue = this.dashboard.getParameterValue(this.parameter),
               value = undefined;
 
-          if(parameterValue.length == 1) {
+          // If no value was specified use current date
+          if (!parameterValue.length){
+            parameterValue = this._getFormattedDate(new Date());
+          } else if(parameterValue.length == 1) {
             parameterValue = parameterValue[0];
           }
 
+          // Get the date well formatted according to the date format specified to pass on to the parser.
+          // For backwards compatibility we need to check if the component is a legacy version of the 
+          // date picker and update the date format accordingly.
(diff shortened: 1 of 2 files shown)


plagoa:
@diogofscmariano @annapopova1 @krivera-pentaho @scottyaslan @rfellows @pamval please review

package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
```diff
@@ -89,16 +89,21 @@ define(['cdf/components/BaseComponent', "dojo/date/locale", 'dijit/form/DateText
           var parameterValue = this.dashboard.getParameterValue(this.parameter),
               value = undefined;
 
-          if(parameterValue.length == 1) {
+          // If no value was specified use current date
+          if (!parameterValue.length){
```
krivera-pentaho: Why wouldn't you just use  [[ if (!parameterValue) { ]]. Is or can parameterValue be null?

plagoa: parameterValue wil return an empty array if no value is specified when creating the prompt, so I have to check for its length.

annapopova1:
Is really needed "_isLegacyDateFormat" and "_convertFormat" functions?
Sorry, but I can't understand why we compare jquery and dojo formatting if now we are using dojo implementation...

pamval:
That was introduced for backward compatibility. If an user opens a dashboard with date picker prompts saved in an earlier version of dashboards (5.4), that logic makes sure the format doesn't break and the dashboard is still usable.

At least that was the goal

annapopova1:
Probably would be useful add comments to code about it. It helps do not forget about this fact for different developers.

pamval:
@plagoa, can you add the comments Anna is mentioning ? I agree it makes a lot of sense to have them.

plagoa:
Updated documentation and made a minor change to the code.

@diogofscmariano @annapopova1 @krivera-pentaho @scottyaslan @rfellows @pamval please review

PR #479, opened 2015-08-18
[BACKLOG-4070] -- enhancement to lifecycle where update calls are mad…

krivera-pentaho:
…e only when the data is changed inside of the component


krivera-pentaho:
@rfellows @scottyaslan @annapopova1 @pamval @plagoa @diogofscmariano please review

PR #480, opened 2015-08-19
[BACKLOG-4070] -- updated _change function to call minimal component …

krivera-pentaho:
…updates and respect 'autoSubmit' checkbox


krivera-pentaho:
@pamval @diogofscmariano @scottyaslan @rfellows @annapopova1 @plagoa please review

PR #484, opened 2015-08-24
[BACKLOG-4823] PIR: unable to edit prompts until a a change is made w…

annapopova1:
…hen opening a report

Details: prompts were not updated when changing edit mode of PIR (because xml is not changed).
To fix this case I added one boolean input parameter to promptPanel.refreshPrompt function. In PIR set it true when change edit mode. And if this parameter is true we will update components.
Also prepared fix for PIR https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/266

@plagoa @diogofscmariano @rfellows @krivera-pentaho @scottyaslan @pamval please review

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -645,12 +645,15 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          * If the new parameter definition is undefined (default impl) no re-initialization will be done.
          *
          * @name PromptPanel#refreshPrompt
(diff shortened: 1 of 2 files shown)


diogofscmariano:
@annapopova1 sounds good, although, i'm thinking why the lifecycle change caused this. do we need to update unit tests due to this?

annapopova1:
@diogofscmariano Before components are always re-created with necessary css. So after change life cycle we got this issue.
Tests ran successful. But we have not enough coverage for "init" function. I will add tests for it.

annapopova1:
@diogofscmariano added some unit tests for "init" function.

PR #486, opened 2015-08-25
[BACKLOG-4038]

diogofscmariano:
- Added logic to avoid need of updating the prompts if the parameter from CDF already has the value selected
 - Removed update of all components if valuesArray of a single component have changed
 - Fixed issue in report viewer that was not removing the error labels after the user gets the input defined

@pamval @krivera-pentaho @annapopova1 @rfellows @scottyaslan @plagoa please review
Dont merged it yet. Tomorrow i want to do more tests and check every unit test. But the implementation should be finished

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -882,14 +882,23 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
                 });
                 if (!isExist) {
                   var errIndex = panel.components.length - 1;
-                  panel.components.splice(errIndex, 0, _createWidgetForErrorLabel.call(this, param, error));
+                  var errorComponent = _createWidgetForErrorLabel.call(this, param, error);
+                  this.dashboard.addComponent(errorComponent);
+                  panel.components.splice(errIndex, 0, errorComponent);
                 }
               }
             }
 
             // checks existing errors components to set correct css style
-            var existingErrorsComponents = _findErrorComponents.call(this, panel);
-            panel.cssClass = (existingErrorsComponents.length > 0) ? (panel.cssClass || '') + ' error' : (panel.cssClass || '').replace(' error', '');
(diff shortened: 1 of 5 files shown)


diogofscmariano:
@pamval @krivera-pentaho @annapopova1 @rfellows @scottyaslan @plagoa pr updated please review

diogofscmariano:
@pamval @krivera-pentaho @annapopova1 @rfellows @scottyaslan @plagoa pr updated please review

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -882,14 +882,21 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
...
+                  panel.components.splice(errIndex, 0, errorComponent);
                 }
               }
             }
 
             // checks existing errors components to set correct css style
             var existingErrorsComponents = _findErrorComponents.call(this, panel);
-            panel.cssClass = (existingErrorsComponents.length > 0) ? (panel.cssClass || '') + ' error' : (panel.cssClass || '').replace(' error', '');
+            if(existingErrorsComponents.length > 0) {
+              panel.cssClass = (panel.cssClass || '') + ' error';
```
annapopova1: Also need check that panel.cssClass does not contain "error" class already. Otherwise we have cssClass with many "error" sub-strings (for example, 'parameter edit error error'). And in the next statement will be replaced only one 'error' sub-string and we will see red border on the page.

I propose we can add checking something like this:
```
            if(existingErrorsComponents.length > 0) {
              if (!panel.cssClass || (panel.cssClass && panel.cssClass.indexOf('error') == -1)) {
                panel.cssClass = (panel.cssClass || '') + ' error';
              }
            } else {
              panel.cssClass = (panel.cssClass || '').replace(' error', '');
              panel.removeErrorClass();
            }
```

diogofscmariano:
@pamval @krivera-pentaho @annapopova1 @rfellows @scottyaslan @plagoa pr updated based on last reviews please review

annapopova1:
Looks good for me

PR #491, opened 2015-08-31
[BACKLOG-4920] -- reverted functionality of selecting a null value ba…

krivera-pentaho:
…ck to parity of 5.4 where when selecting a null value, it reverts back to the most recently selected value


krivera-pentaho:
@pamval @diogofscmariano @annapopova1 @plagoa @scottyaslan

PR #492, opened 2015-09-01
[BACKLOG-4916] Running Prpt sample, report does not render correctly …

annapopova1:
…[regression]

We should initialize all parameters, also with hidden attribute. Because hidden parameters can be used by plugins. For example, "htmlProportionalWidth" parameter is used by Report Viewer to choose concrete css class.

@plagoa @diogofscmariano @rfellows @krivera-pentaho @scottyaslan @pamval please review


diogofscmariano:
@annapopova1  Please add unit tests to it

PR #497, opened 2015-09-04
[BACKLOG-5059]

diogofscmariano:
- Fixed issue with editing date picker due to missing property about the digit id
- Reviewed date format handing due to double conversions and non-conversions of some legacy formats
- Handled special case for y and yy in dojo due to automatic conversions to yyyy
- Updated unit tests accordingly

@annapopova1 @plagoa @pamval @krivera-pentaho @scottyaslan please review


krivera-pentaho:
Looks good to me.

PR #504, opened 2015-09-18
[BACKLOG-5149] - prpt Date Picker Prompt does not behave as expected …

krivera-pentaho:
…in PUC

- Added specific css file to the DojoDateTextBoxComponent with specific overrides to global styles;


krivera-pentaho:
@pamval @dkincade this is the cherry pick of @plagoa's pull request from earlier.

PR #509, opened 2015-09-24
[BACKLOG-5307] PIR: Error message - This prompt is mandatory - appear…

krivera-pentaho:
…s for prompt with selected value


krivera-pentaho:
@dkincade please review

PR #511, opened 2015-09-25
[BACKLOG-5307] - PIR: Error message "This prompt is mandatory" appear…

plagoa:
…s for prompt with selected value

- Added timeout set to zero to allow for error message to be added to the DOM by the time the lifecycle is clearing it;


plagoa:
Cherry pick to 6.0

@diogofscmariano @krivera-pentaho @annapopova1 @scottyaslan @pamval @dkincade please review

PR #516, opened 2015-09-28
[BACKLOG-5355] -- cleared nullValueParams variable to halt infinite loop

krivera-pentaho:


krivera-pentaho:
@dkincade please review
already tested and merged into master. unit tests are passing.

krivera-pentaho:
@pamval please review

dkincade:
But were the unit tests passing before? (can we add a test to prevent this from regressing in the future)?

krivera-pentaho:
Unfortunately not. This would have to be an integration test that we would have to add to verify that this problem won't happen again. It was spanning multiple "on" events which we would not be able to capture inside of a unit test.

PR #521, opened 2015-10-20
[BISERVER-12644][COMP-1665] Update requirejs to version 2.1.19

dcleao:
This PR should only be accepted if pentaho/subfloor#24 is accepted.

* Update require.js to version 2.1.19
* Fix dojo 1.9 to support new requireJS (https://bugs.dojotoolkit.org/ticket/17829)
* Moved declaration of "pen.require" and "pen.define" to "require-cfg.js", which is loaded immediately after "require.js" by "webcontext.js".


dcleao:
@pentaho-nbaker, @lgrill-pentaho, @pamval please review.

package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/kernel.js
```diff
@@ -10,6 +12,7 @@ define(["../has", "./config", "require", "module"], function(has, config, requir
 
 		// create dojo, dijit, and dojox
 		// FIXME: in 2.0 remove dijit, dojox being created by dojo
+		global = (function () { return this; })(),
 		dijit = {},
```
pentaho-nbaker: I remember removing this line. There There RequireJS case on it, but "this" isn't window inside RequireJS callbacks

package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/kernel.js
```diff
@@ -18,7 +21,7 @@ define(["../has", "./config", "require", "module"], function(has, config, requir
 
 			// notice dojo takes ownership of the value of the config module
 			config:config,
-			global:window,
+			global:global,
 			dijit:dijit,
```
pentaho-nbaker: Same. I did this on purpose

package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/kernel.js
dcleao: Using a global variable determined like this guards us against exactly that problem!

package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/kernel.js
dcleao: `window` is not defined in node.js, while global has the correct value in node.js as well.

package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/kernel.js
dcleao: This version of require.js changed the value of `this` when calling the module definition function, from the global object to the module's exports object, so uses of `this` directly in the definition function broke. But not uses of this in other functions ;-) as is the case of the determination of the `global` variable.

pentaho-nbaker: Ha! You're right. The anonymous function is perfect.

PR #523, opened 2015-11-11
[ENGOPS-1578] - Updated cdf js artefact

diogofscmariano:


diogofscmariano:
@lgrill-pentaho please review

PR #525, opened 2015-11-16
PIR-1192 - Interactive Reports does not show fields with datasource n…

Aliaksandr-Kastenka:
…ame has Japanese characters


Aliaksandr-Kastenka:
@pamval 
Please review.

pamval:
@rfellows can you please take a look at this one ?
My comment is: Can we unit test this ? I know we have the harness in place for common-ui.

rfellows:
@pamval It "can" be tested, like anything. However, IMO if you want it unit tested the whole method should be refactored to make testing easier. In particular, the pentahoGet call should be replaced with the equivalent dojo function. The code that was changed doesn't introduce any risk that i would think would warrant such a refactor. So, it's up to you if you want Aliaksandr to spend time doing that or not.

PR #529, opened 2015-11-25
[BACKLOG-5541] -- Updated api.operation.render method to accept both …

krivera-pentaho:
…a synchronous and asynchronous method for getParameterXml

### DIFF
--- a/package-res/resources/web/prompting/api/OperationAPI.js
+++ b/package-res/resources/web/prompting/api/OperationAPI.js
@@ -48,39 +48,55 @@ define(['common-ui/prompting/PromptPanel', 'common-ui/prompting/parameters/Param
      *
      * @name OperationAPI#render
      * @method
-     * @param {String} id HTML object id where the render the panel
-     * @param {Function} getParameterXml The function called when the prompt panel needs to get the new parameter xml
+     * @param {String}    id HTML object id where the render the panel
+     * @param {Function}  getParameterXml The function called when the prompt panel needs to get the new parameter xml.
+     *                                    Receives a callback function which needs to be executed either synchronously
+     *                                    or asynchronously.
+     * @example
+     *  // Asynchronous
+     *  api.operation.render("prompt-panel-render-area", function(api, callback) {
+     *    retrieveXml(function(xml){
+     *      callback(xml);
+     *    });
+     *
+     *  // Synchronous
+     *  api.operation.render("prompt-panel-render-area", function(api, callback) {
+     *    var xml = retrieveXml();
+     *    callback(xml); // return already retrieved xml definition
+     *  });
      */
     this.render = function(id, getParameterXml) {
       if (!getParameterXml || typeof getParameterXml != "function") {
         api.log.error(this._msgs.NO_PARAM_XML_FUNC, true);
       }
 
-      var xml = getParameterXml(api);
-      if (!xml) {
-        api.log.error(this._msgs.NO_PARAM_XML, true);
-      }
-
-      var paramDefn = this._parameterParser.parseParameterXml(xml);
-      this._promptPanel = new PromptPanel(id, paramDefn);
-
-      // Override of getParameterDefinition
-      this._promptPanel.getParameterDefinition = (function(promptPanel, refreshCallback) {
-        var xml = getParameterXml(api);
+      getParameterXml(api, (function(xml) {
         if (!xml) {
-          api.log.error(this._msgs.NO_PARAM_XML);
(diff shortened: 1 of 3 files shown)


krivera-pentaho:
@diogofscmariano @pamval @annapopova1 please review

annapopova1:
The first implementation provided only the synchronous way. Therefore IMO better use the synchronous way by default. It also better for users who already use this API. What do you think guys?

diogofscmariano:
The current implementation is not an issue. The API is not public and we can follow an approach that rewrites it if needed. That is the reason why the API was not published on 6.0.

More global question here... if we need to provide a async and sync way of rendering a prompt why we don't think this from the opposing way? For instance, no matter of what is the way getParameterXml is called (either async or sync), we must provide a much cleaner way for the users to do it. I don't like exposing that info to the user in the method. Why we don't go with something like config the call like jQuery does, via a config on the api or using a json? Also, i don't see the user switching from sync to async. About default or not, IMO we should have the default as async. @pamval @annapopova1 @plagoa @krivera-pentaho  what you guys think?

pamval:
Why don't we just assume the getParameterXML always receives a callback and leave the sync/async definition to whoever implements that function ?

That way, you don't need a boolean on the signature nor a config.

krivera-pentaho:
Sounds good. I will make the changes so that it always receives a callback. If a value is returned, then it is assumed synchronous, otherwise it is asynchronous

krivera-pentaho:
@diogofscmariano @annapopova1 @plagoa @pamval, i updated the PR. please review

package-res/resources/web/prompting/api/OperationAPI.js
```diff
@@ -48,39 +48,62 @@ define(['common-ui/prompting/PromptPanel', 'common-ui/prompting/parameters/Param
      *
      * @name OperationAPI#render
      * @method
-     * @param {String} id HTML object id where the render the panel
-     * @param {Function} getParameterXml The function called when the prompt panel needs to get the new parameter xml
+     * @param {String}    id HTML object id where the render the panel
+     * @param {Function}  getParameterXml The function called when the prompt panel needs to get the new parameter xml.
+     *                                    Receives a callback function. If synchronous, simply return xml. If asynchronous,
+     *                                    call the callback function provided and and pass xml to that.
```
annapopova1: double "and"

package-res/resources/web/prompting/api/OperationAPI.js
```diff
@@ -48,39 +48,62 @@ define(['common-ui/prompting/PromptPanel', 'common-ui/prompting/parameters/Param
...
-      var xml = getParameterXml(api);
-      if (!xml) {
-        api.log.error(this._msgs.NO_PARAM_XML, true);
-      }
-
-      var paramDefn = this._parameterParser.parseParameterXml(xml);
-      this._promptPanel = new PromptPanel(id, paramDefn);
+      var getParameterXmlAsync = function(callback) {
+        var xml = getParameterXml(api, callback);
+        if (typeof xml != "undefined") {
```
annapopova1: probably unnecessary check because "xml" is checked into callback (row 82).

krivera-pentaho: I check specifically for "undefined" here as an indication of whether we are trying to have a synchronous or asynchronous function call. if (!xml) will be for null, or other falsy values.

krivera-pentaho:
@diogofscmariano @pamval @annapopova1 @plagoa please review. I removed the 2nd "and"

package-res/resources/web/prompting/api/OperationAPI.js
annapopova1: @krivera-pentaho thank you, I understood. I have not more comments.

pamval:
So, what I meant before is that there should be no distinction (from inside the Prompting API) if the handler is synchronous or asynchronous.

That means that we always expect the implementer of the retrieveParameters function to call the callback after the parameter XML has been retrieved. Regardless of whether that retrieval was sync or async.

I'm not sure if I'm using the correct highly technical terms here, @krivera-pentaho , so be patient.

krivera-pentaho:
LOL!!!!! It was too highly technical for me to understand ;)

I understand. So you don't even want the function to even be able to return anything. Always call the callback function. They either call it synchronously or asynchronously, within their implementation for getParameterXml. I will change the implementation.

krivera-pentaho:
@diogofscmariano @annapopova1 @pamval @plagoa please review

PR #530, opened 2015-11-26
[BACKLOG-5539, TEST] - Some prompting files are assuming the existenc…

plagoa:
…e of a jquery global, inside of AMD injected

- Removed dependency on globally defined jQuery;
- Added new require configuration for a clean jQuery used by prompting;
- Updated the builders and components modules to require jQuery in this fashion;
- Updated both samples pages (regular and API version);
- Updated the tests;


plagoa:
@diogofscmariano @krivera-pentaho @annapopova1 @pamval please review

PR #531, opened 2015-11-27
[TEST] - Some prompting files are assuming the existence of a jquery …

plagoa:
…global, inside of AMD injected

- Fixed function name on test;


plagoa:
@diogofscmariano please review

PR #532, opened 2015-12-02
[ANALYZER-2893] - Long measure names are not rendered correctly on Analyzer Layout pane

akhayrutdinov:
- crop long names

@rfellows, @bennychow, @krivera-pentaho, review it please

### DIFF
--- a/package-res/resources/web/dojo/dijit/themes/pentaho/layout/LayoutPanel.css
+++ b/package-res/resources/web/dojo/dijit/themes/pentaho/layout/LayoutPanel.css
@@ -94,6 +94,8 @@
   height: 18px;
   line-height: 17px;
   display: inline-block;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .gem .gemMenuHandle{


krivera-pentaho:
Did we also want to truncate the end and add a "..." or just hide the overflow?

akhayrutdinov:
@krivera-pentaho, that's a good question!
Actually, I decided to keep the behaviour of `gem-label` the same, as pivot's table header has (it does not suffer from the problem descibed in the ticket). The header has `text-overflow: clip` (default), meaning it is simply cutting the extra text. Also, we have left panel with controls and long text is also simply truncated there.

lucboudreau:
Please make sure someone from UX pitches in on this one.

akhayrutdinov:
OK, I've sent a letter to Wes (you are CCed). Let's wait for the UX feedback

akhayrutdinov:
@krivera-pentaho, I've updated the PR according to UX team's recommendations. Review it please. There'll be another PR for Analyzer

PR #533, opened 2015-12-09
BISERVER-12961 - Setting drop-down parameter causes text box to revert…

Aliaksandr-Kastenka:
… to default values

When drop-down parameter array formatted
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L962
extra value (emty) added and this check
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L969
is passing, which leads to update with default values.

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -966,7 +966,15 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
 
                 // Compare values array from param (which is formatted into valuesArray) with the current valuesArray
                 // We need to update the components if autoSubmit is off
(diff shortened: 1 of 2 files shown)


Aliaksandr-Kastenka:
@diogofscmariano @pamval 
Please review.

pamval:
@plagoa @krivera-pentaho @annapopova1 please review this

annapopova1:
I think anyway we need write js unit tests for this situation.

krivera-pentaho:
it looks good, but i agree, there should be unit tests.

Aliaksandr-Kastenka:
@annapopova1 @krivera-pentaho 
Please review.

annapopova1:
It's good for me.

PR #534, opened 2015-12-10
[BISERVER-12958] Date Picker Parameters default to Today's date in pr…

ivanpogodin:
…pt reports when running on BA Server
@pentaho-nbaker , @rfellows , @mdamour1976 , could you review it, please?
Fix: DojoDateTextBoxComponent default value is null (instead of now).
Now the DateTextBox default value is null wherever this component is used.


rfellows:
@krivera-pentaho @scottyaslan @diogofscmariano please review this. it's in the prompting area. thought you might have a better idea if this might cause any issues or not.

krivera-pentaho:
Actually, @plagoa and @annapopova1 should take a closer look at this one.

annapopova1:
Looks like it's correct.
FYI, this component is used also in PIR and Dashboard plugins. So in 5.4 datepicker in Dashboard plugin by default also does not show current date but shows empty.

PR #544, opened 2015-12-22
[BACKLOG-5780] - Making sure dateFormats with mm are recognized as legacy

jvelasques:


jvelasques:
@pamval @diogofscmariano @plagoa @annapopova1 please review

package-res/resources/web/prompting/components/DojoDateTextBoxComponent.js
```diff
@@ -213,6 +213,10 @@ define(['cdf/components/BaseComponent', "dojo/date/locale", 'dijit/form/DateText
             return true;
           }
 
+          if(dateFormat.match(/mm/)) {
+            return true;
+          }
+
```
pamval: Isn't this too generic ? I thought we were going to focus on the specific yy-mm-dd case. 
Doesn't this break use cases where we are trying to show minutes instead of months ?

diogofscmariano: dont think fixing it to yy-mm-dd is the right solution. Things like y-mm-dd, yy-m-d are also formats that in dojo date picker dont make sense because they will be picking minutes. I think if we find the existence of the m string on our use case is probably the best solution to cover all the cases. If we mark this as legacy, then the current conversion method will deal with it

pamval: That means we're never showing time (which is probably reasonable on our current date picker). ok, then.

jvelasques:
@pamval @diogofscmariano incorporated feedback please review

PR #548, opened 2015-12-22
[BACKLOG-5542] -- updated PromptPanel to not require a parameter defi…

krivera-pentaho:
…nition to be created. Updated API for these changes. Updated unit tests accordingly

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -368,12 +368,7 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
            */
           this.destinationId = destinationId;
 
-          if (!paramDefn) {
-            throw 'paramDefn is required';
-          }
-          this.paramDefn = paramDefn;
-
-          this.autoSubmit = paramDefn.allowAutoSubmit();
+          this.setParamDefn(paramDefn);
 
           this.promptGUIDHelper = new GUIDHelper();
 
@@ -384,6 +379,30 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
           this.paramDiffer = new ParamDiff();
 
           this.widgetBuilder = WidgetBuilder;
+        },
+
+        /**
+         * Returns the parameter definition if it has been set. Otherwise an exception is thrown.
+         *
+         * @returns {Object}
+         */
+        getParamDefn: function() {
+          if (!this.paramDefn) {
+            throw 'paramDefn is required. Call setParameterDefn';
+          }
+
+          return this.paramDefn;
+        },
+
+        /**
+         * Sets the parameter definition for the prompt panel. Also sets whether the prompt panel has auto submit
+         * @param paramDefn {Object} The parameter definition object
+         */
+        setParamDefn: function(paramDefn) {
+          this.paramDefn = paramDefn;
+          if (paramDefn) {
+            this.autoSubmit = paramDefn.allowAutoSubmit();
+          }
         },
 
         /**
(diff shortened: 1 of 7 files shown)


krivera-pentaho:
@diogofscmariano @annapopova1 @plagoa @pamval @tmorgner please review. 

@tmorgner I made some api changes because prompt panels do not need a parameter definition. To create a new API, an HTML id is required. Also, the operation.render method does not require that HTML id. The operation API creates a prompt panel when it is required and created by the parent API object.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -361,6 +360,19 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
           this.widgetBuilder = WidgetBuilder;
         },
 
+        getParamDefn: function() {
+          if (!this.paramDefn) {
+            throw 'paramDefn is required. Call setParameterDefinition';
```
annapopova1: 1. 'setParameterDefinition' - is not exist. Do you mean PromptPanel#setParamDefn?
2. Need add js doc to 'getParamDefn' and 'setParamDefn' functions

package-res/resources/web/prompting/api/OperationAPI.js
```diff
@@ -54,18 +54,18 @@ define(['common-ui/prompting/PromptPanel', 'common-ui/prompting/parameters/Param
...
      *
      *  // Synchronous
-     *  api.operation.render("prompt-panel-render-area", function(api, callback) {
+     *  api.operation.render(function(api, callback) {
      *    var xml = retrieveXml();
      *    callback(xml); // return already retrieved xml definition
      *  });
      */
-    this.render = function(id, getParameterXml) {
+    this.render = function(getParameterXml) {
```
annapopova1: Need remove unused parameter in js doc: "@param {String}    id HTML object id where the render the panel"

annapopova1:
Also need change prompting\api\promptApiSample.html page because after this changes it will not work.

krivera-pentaho:
@tmorgner I agree and remove the catch.
@annapopova1 I will updatte the promptApiSample, since we are apparently still using it.

krivera-pentaho:
@tmorgner I instead checked which exception string is being thrown and if it matches the one i need, then i log a warning, otherwise i re throw the exception.

@diogofscmariano @annapopova1 @plagoa 
I've addressed the issues. Please review.

package-res/resources/web/prompting/api/OperationAPI.js
```diff
@@ -26,9 +26,9 @@
  * @property {Object} _msgs Contains possible constants of messages
  */
 define(['common-ui/prompting/PromptPanel', 'common-ui/prompting/parameters/ParameterXmlParser'], function(PromptPanel, ParameterXmlParser) {
-  return function(api) {
+  return function(api, id) {
     this._parameterParser = new ParameterXmlParser();
-    this._promptPanel = null;
+    this._promptPanel = new PromptPanel(id);
```
diogofscmariano: if we are building a parameter panel without paramDefn will this https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/548/files#diff-fe97c9bc6920bca0d75fa6ad17207ee6R348 always fail? This check should probably not be here or this will never set the value of the auto submit

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -343,12 +343,11 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
            */
           this.destinationId = destinationId;
 
-          if (!paramDefn) {
-            throw 'paramDefn is required';
-          }
           this.paramDefn = paramDefn;
```
diogofscmariano: even better, lets remove this from here and use the setParamDefn method! actually it already does what we need

diogofscmariano:
apart from this comment, everything looks good

package-res/resources/web/prompting/api/OperationAPI.js
krivera-pentaho: Yes, it will fail, in the case of the API. Other places can still create the prompt panel with a parameterDefinition as part of the constructor.

package-res/resources/web/prompting/PromptPanel.js
krivera-pentaho: I will call setParamDefn.

krivera-pentaho:
@diogofscmariano i have made the changes. Please review.

PR #551, opened 2016-01-06
[BACKLOG-5981] - Common Client-Side Metadata Model - Initial implementation

dcleao:
### UTIL
* AMD namespace `pentaho/util`
* Adapted from old `pentaho/visual/_utils` module
* arg, error, fun, object, text, promise

### LANG
* AMD namespace `pentaho/lang`
* `Base` - updated version of the `Base.js` library (needs tests)
* `List` and `Collection` classes (adapted from old `pentaho/visual/data/Collection` and `.../List`) (needs tests)

### THIN-CLIENT
* `pentaho/i18n` (needs tests)
  * AMD plugin
  * `pentaho/util/MessageBundle` class (with a handy `structured` property, with the compliments of Nelson Antunes)
  * i18nMock for testing modules which use the i18n AMD plugin
  * `pentaho/common/Messages` added functionality to the `Messages#addBundle` method
     to allow registration of prefetched bundles

* `pentaho/service`
  * Moved to "pentaho" folder
  * Now using `pentaho/util`

### SHIM
* `es6-promise-shim` renamed to `pentaho/shim/es6-promise`
* `es5-shim` renamed to `pentaho/shim/es5`

### DATA
* AMD namespace `pentaho/data`
* Promoted VizAPI 2.5 data namespace, `pentaho/visual/data`, to `pentaho/data`
* Now using `pentaho/util` and `pentaho/lang`
* Docs still in YUIDocs format...

### METADATA MODEL
* AMD namespace `pentaho/type`
* Class `Context`
  * Context variables: container, user, theme and locale
  * Each context holds differently configured metadata models
* Base classes `Item` / `Item.Meta`
   * A type is composed of two classes, the "mesa" (`Item`) and its "meta" (`Item.Meta`) counterpart
   * Defining the "mesa" class, simultaneously defines the corresponding "mesa" class
   * The syntaxes `Item.extend({meta: .}, {meta: .})`, `Item.implement({meta: .})` and `Item.implementStatic({meta: .})`
     allow configuring the `Meta` class
   * `Context` independent
* Classes `Value`, `Simple` and `Complex` (and their corresponding .Meta classes)
   * Contextful - each context has a distinct "version" of these
* Classes `Property` and `PropertyMetaCollection` for complex types
* Initial implementation of dynamic attributes of Properties
* Partially complete JsDocs and Tests

### VizAPI 3.0
* Moved Experimental VizAPI 2.5 to `web/pentaho/visual` folder
* Added a sample viz `pentaho/visual/samples/calcNew` based on the _future_ 3.0 API

### DOC
* Updated to jsdocs 3.4.0, for the new `useLongnameInNac` default template option

### OTHER
* Fixed javascript to cope with PhantomJS 1.9.8's weak ES5 support

@pamval, @graimundo, @scottyaslan, @davidmsantos90, @nantunes, @marcovala, @carlosrusso please provide your valuable feedback.

Do *NOT* merge this until https://github.com/pentaho/pentaho-analyzer/pull/899 can be merged as well.


nantunes:
Does it make sense to move es6-promise (or any other external code) to the pentaho namespace?

dcleao:
@nantunes The rationale is that these names are not "universal" — the `es6-promise` AMD id was a choice of ours (although it does coincide with the original file name). The main reason is ensuring that our solutions work in embedding scenarios, where we don't own the AMD system, and thus global AMD ids.
The case of 3rd party libs (like this one) is a gray one. In the case of shims, we have also the "es5-shim", which is not a complete ES5 shim, but just a handful of `Array` methods. Thus, registering at the global "es5-shim" name would not be correct. This one was also moved, along with `es6-promise-shim`, to `pentaho/shim/es5`.

Not the case of the "es6-promise" lib though, which is complete and from a 3rd-party provider. At the current state of the platform's AMD system, I'd see this as _the version of "es6-promise" that is shipped with Pentaho 6.1_. When our platform's AMD system gets smarter wrt to versions, we can turn this into an "alias" of a versioned module in its producer's namespace.

There is one other thing. The `pentaho/shim/es6-promise` is not directly the producer's file, but one that installs the shim conditionally when promises are not detected. The producer's file is "privately" available in `pentaho/shim/_es6-promise/es6-promise`.

dcleao:
ping @pamval, @graimundo, @scottyaslan, @davidmsantos90, @marcovala, @carlosrusso.

nantunes:
@dcleao On it right now. ;-)

nantunes:
The calcNew sample must be updated to use property "view" instead of "viewId" (typeFactory.js).

The value of that property is supposably evaluated as relative to the declaring type. However it doesn't take into account the use of ```packages``` and its ```main``` file, so it is really using the parent folder as base path.

I also think that this shouldn't be merged without @carlosrusso 's ```pentaho/visual``` class.

Will update if I find anything more.

dcleao:
@nantunes Thanks for noting those fixes. If I'm not mistaken, those things have already been fixed (or will be) on the group of changes following this one, either on the [BACKLOG-6050](http://jira.pentaho.com/browse/BACKLOG-6050) or in [BACKLOG-6020](http://jira.pentaho.com/browse/BACKLOG-6020).
The PR explicitly says this is the _initial_ implementation, because it is known that there are some open ends :-) .
Do you agree?

graimundo:
@dcleao, am I correct by saying that besides the impact on Analizer that is resolved with https://github.com/pentaho/pentaho-analyzer/pull/899, there are no other foreseeable implications?  If so I would say we should go ahead and merge this PR along with the others that will resolve the issues @nantunes mentioned.

nantunes:
@graimundo Just irrelevant side note: the DET POC is affected by the changes in the service plugin.

dcleao:
@graimundo yes. However, I wonder if we should be merging this to master now.

1. This new code will be in 6.1
2. CDF still uses VizAPI 2.5
3. This PR does not break 2.5, but until 6.1 is out, we should have the need/chance to start using the pentaho/visual namespace for 3.0.
4. So we might have to create a new branch for some of the 3.0 code until 6.1 is out...

What do you think? @pamval any thoughts on this?

dcleao:
@nantunes good point :-)

pamval:
I think this should be merged into master now.

If we need to branch later, we will.

PR #552, opened 2016-01-11
[BACKLOG-5840] -- added ability to set block ui options through the p…

krivera-pentaho:
…rompt api

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -382,6 +382,15 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
         },
 
         /**
+         * Returns the dashboard store on the prompt panel
+         *
+         * @return {Object}
+         */
+        getDashboard: function() {
+          return this.dashboard;
+        },
+
+        /**
          * Returns the parameter definition if it has been set. Otherwise an exception is thrown.
          *
          * @returns {Object}
@@ -1336,7 +1345,7 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          * @method
          */
         showProgressIndicator: function() {
-          this.dashboard.showProgressIndicator();
+          this.getDashboard().showProgressIndicator();
         },
 
         /**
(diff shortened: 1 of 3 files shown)


krivera-pentaho:
Requires https://github.com/webdetails/cdf/pull/744 to be merged
@diogofscmariano @jvelasques @plagoa @annapopova1 @pamval please review

package-res/resources/web/prompting/api/UiAPI.js
```diff
@@ -51,5 +51,40 @@ define([], function() {
...
+     *              backgroundColor : '#000000',
+     *              opacity : '0.6',
+     *              cursor : 'wait'
+     *          },
+     *          showOverlay : false
+     *      };
+     *      api.ui.setBlockUiOptions(defaults);
+     */
+    this.setBlockUiOptions = function(options) {
+      api.operation._getPromptPanel().dashboard._setBlockUiOptions(options);
```
pamval: The API should not access the dashboard in prompt panel directly. The dashboard should be a private member of the prompt panel.

krivera-pentaho: I will fix this in the prompt panel

krivera-pentaho:
@diogofscmariano @pamval @jvelasques @plagoa @annapopova1 please review.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -382,6 +382,15 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
...
 
         /**
+         * Returns the dashboard store on the prompt panel
+         *
+         * @return {Object}
+         */
+        getDashboard: function() {
+          return this.dashboard;
+        },
+
```
annapopova1: If we have getDashboard, probably need re implement showProgressIndicator and hideProgressIndicator from PromptPanel because before that we encapsulated dashboard functions into PromptPanel.

pamval: We should encapsulate the setBlockUIOptions in the promptpanel as well. 
dashboard "feels" like a internal to prompt panel. I don't think it should be exposed.

diogofscmariano: +1 here

krivera-pentaho: I am not disagreeing, but i wonder why we try to encapsulate so much inner functionality from the API? If we make the API a "passthrough" to encapsulated functions instead of providing implementations, then why wouldn't we just document and expose the prompt panel object's public functions?

krivera-pentaho:
@diogofscmariano @plagoa @jvelasques @annapopova1 @pamval please review

pamval:
that's a good point. I think that the need for this separation will become clearer once we get to implementing alternative prompt providers. 

IMO, choosing and instantiating the correct prompt provider will be done at the API level (maybe fetching the correct implementation for the PromptPanel).

That way, we keep the interface to the api clients exactly the same and they don't have to bother with the mechanics of choosing a PromptProvider, they ask the API to do it.

And this means that the prompt api -> Prompt panel interface must also be very clean and not access internal members since we'll maybe need to switch PromptPanel implementations (although accessing internal members from the outside is a no-no in any scenario).

Anyway, we should discuss this further. Let's schedule a meeting for after feature freeze. Until then, I suggest we keep the approach we've been following so far.

PR #554, opened 2016-01-14
[BACKLOG-5844] Prompt API - Change api.operation.render so callback "…

annapopova1:
…getParameterXML" will return parsed ParameterDefinition.

- changed render function to apply ParameterDefinition
- changed unit tests, also add tests for init function
- changed promptApiSample.html. Probably need remove it if we do not want support it.

@krivera-pentaho @pamval @diogofscmariano @plagoa @afrjorge @jvelasques  please review


annapopova1:
Also added api.util.parseParameterXml function as separate commit. Could you take a look?

PR #565, opened 2016-01-19
[BACKLOG-6020] Proof-of-concept implementation of base Model and View classes

carlosrusso:


carlosrusso:
This pull request should only be merged after #564

PR #549, opened 2015-12-29
[BISERVER-12918] - the report parameter/prompt panel flashes when auto submit off

akhayrutdinov:
@pamval, @diogofscmariano, @krivera-pentaho, review it please. This is a replacement of https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/546

@MikhailHubanau is absent until the middle of Jan, that is why I picked his changes and added tests as well as some small changes.

PR #568, opened 2016-01-21
[SP-2416] - Backport of BISERVER-12918 - the report parameter/prompt panel flashes when auto submit off (6.0 Suite)

akhayrutdinov:
- add _areParamsDifferent() as an additional check for values equality
- change _changeComponentsByDiff() logic: do not call _areParamsDifferent() if it is clear, that update is needed
- add tests for new changes
(adopted from commit b996eb5)

@diogofscmariano, review it please. This is a backport of https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/549


diogofscmariano:
@pamval this is a simple backport. this sounds good to me

PR #569, opened 2016-01-21
[BACKLOG-5981] - Common Client-Side Metadata Model

dcleao:
* Split the `Value` class into `Element` and `List` classes. `Complex` and `Simple` now extend `Element`.
  * pentaho/type/list - code, docs and unit tests.
  * Properties do not have a settable "list" attribute anymore.
    It is now derived from whether the property type is of a list type or not.
* Added Value#key property and implemented specially: Simple#key, Complex#uid, Complex#key
* Complex
  * Fixed inconsistencies among Complex#{get, get1, getv, getf} in what throwing errors is concerned
    * Changed the methods with signatures like `get(prop, assertDefined)` into the safe-by-default `get(prop, lenient)`
  * Added `Complex#path` method to allow getting multiple steps in a null-safe way; includes unit-tests
  * Added unit tests for Complex#count(name)
* Implemented Context#create code; unit tested.
* Assorted
  * Added `phantomjs-shims.js` file to test environment to shim Function#bind; this is only surely defined when certain Specs are loaded...
  * Fixed a PhantomJS problem in Value#context returning a variable named context as well...
  * Changed the regexps on `karma.*.conf.js` files, in the files/pattern properties, because of a bug with more recent versions of karma.
  * Converting `a instanceof Array` expressions to `Array.isArray`
  * Moved isSubsetOf function to the new utils/array module.

@pamval, @graimundo, @scottyaslan, @davidmsantos90, @nantunes, @marcovala, @carlosrusso please provide your valuable feedback.

As it is now, it is good enough for you guys to base future changes. However, I will be adding another commit soon with some more unit tests, to cover some of the new stuff I'm adding that is lacking coverage.


package-res/resources/web/pentaho/type/simple.js
```diff
@@ -149,11 +175,42 @@ define([
...
+       * If two values of the same concrete type have distinct keys,
+       * then {@link pentaho.type.Value.Meta#areEqual} should return `false`.
+       *
+       * The default simple value implementation, returns the result of calling
+       * `toString()` on {@link pentaho.type.Simple#value}.
+       *
+       * @type string
+       * @readonly
+       */
+      get key() {
```
nantunes: This don't work as intended with ```pentaho/type/object```. The ```toString``` always return ```"[object Object]"```, so it is always considered equal.

Maybe in this case we could probably just consider equals if it indeed is the same instance. That would need an override of the ```equals``` function by ```pentaho/type/object``` so that don't check the ```key``` property. The doc description of this getter would also have to change to break the hard requirement between the two functions.

Other option is to override the ```key``` getter in ```pentaho/type/object``` so it returns an hash of the object (i.e., ```md5``` of a ```JSON.stringify```). It would become, however, a much costly function.

package-res/resources/web/pentaho/visual/base/modelFactory.js
```diff
@@ -48,21 +48,8 @@ define([
      * @param {pentaho.visual.base.IModel} modelSpec A plain object containing the model specification.
      */
     var Model = Complex.extend({
-        constructor: function(spec) {
-          this.base(spec);
-          var missingProps = this.meta.props.filter(function(prop) {
```
nantunes: ```props``` no longer exists.
(fixed now, comment just for reference)

PR #571, opened 2016-01-22
[BACKLOG-5846] - Extend EventAPI#parameterChanged to subscribe to cha…

plagoa:
…nges in specifc, named parameters

- updated the parameterChanged method on the event api to allow the registration of a callback to the change of a specific parameter;
- updated the promptPanel onParameterChanged property to have an object of "parameter / callback" pairs instead of a function;
- updated the promptPanel parameterChanged method to call the callback for a specific parameter ;
- added tests for the event api;
- added tests for the promptPanel;

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -26,6 +26,7 @@
  * @property {Boolean} autoSubmit True if the prompt is in auto submit mode, false otherwise
  * @property {Dashboard} dashboard The dashboard object assigned to the prompt
  * @property {Boolean} parametersChanged True if the parameters have changed, False otherwise
+ * @property {Object} onParameterChanged collection of parameterNames and the callback called when that parameter is changed
  * @property {Function} onBeforeRender Callback called if defined before any change is performed in the prompt components
  * @property {Function} onAfterRender Callback called if defined after any change is performed in the prompt components
  * @property {Function} onBeforeUpdate Callback called if defined before the prompt update cycle is called
@@ -744,7 +745,14 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          */
         parameterChanged: function (param, name, value) {
           if (this.onParameterChanged) {
-            this.onParameterChanged(name, value);
+            var paramCallback = this.onParameterChanged[name];
+            if (paramCallback) {
+              if (typeof paramCallback === 'function') {
+                paramCallback(name, value); 
+              } else {
+                Logger.warn("The parameterChanged callback for '" + name + "' is not a function");
+              }
+            }
           }
 
           if (!value || value == "" || value == "null") {
--- a/package-res/resources/web/prompting/api/EventAPI.js
+++ b/package-res/resources/web/prompting/api/EventAPI.js
@@ -79,13 +79,31 @@ define([], function() {
      * @method
      * @param {Function} callback   The function to be executed when the event is triggered.
      *                              Pass null if you wish to unbind this event
-     * @example
+     * @param {String} paramName    The name of the parameter to which the callback will be called
+     *                              This argument is optional. If no parameter name is passed on
+     *                              the callback will be assigned to all the parameters. If paramName is
+     *                              not a valid string the callback will be registered to all parameters.
+     *
+     * @example <caption>Register a parameter changed event to all parameters</caption>
      *  api.event.parameterChanged(function(parameterName, parameterValue) {
      *    // Execute event based code
      *  })
+     *
+     * @example <caption>Register a parameter changed event to the parameter called "parameterName"</caption>
+     *  api.event.parameterChanged(function(parameterName, parameterValue) {
+     *    // Execute event based code
+     *  }, "parameterName")
      */
-    this.parameterChanged = function(callback) {
-      api.operation._getPromptPanel().onParameterChanged = callback;
+    this.parameterChanged = function(callback, paramName) {
+      if (callback) {
+        if (!api.operation._getPromptPanel().onParameterChanged) {
+          api.operation._getPromptPanel().onParameterChanged = {};
+        }
+        paramName = (typeof paramName === "string" && paramName) || '';
+        api.operation._getPromptPanel().onParameterChanged[paramName] = callback; 
+      } else {
+        api.operation._getPromptPanel().onParameterChanged = null;
+      }
     };
 
     /**
--- a/package-res/resources/web/test/prompting/PromptPanelSpec.js
+++ b/package-res/resources/web/test/prompting/PromptPanelSpec.js
@@ -354,6 +354,32 @@ define([ 'dojo/number', 'dojo/i18n', 'common-ui/prompting/PromptPanel',
         expect(panel.nullValueParams[0]).toBe(param);
       });
 
+      it("parameterChanged with specific parameter callback", function() {
+        var param = {};
+        var name = "paramName";
+        var parameterChangedSpy = jasmine.createSpy("ParameterChangedSpy");
+        panel.onParameterChanged = {};
+        panel.onParameterChanged[name] = parameterChangedSpy;
+        panel.parameterChanged(param, name);
+        expect(parameterChangedSpy).toHaveBeenCalled();
+      });
+
+      it("parameterChanged with multiple parameter callback", function() {
+        var param = {};
+        panel.onParameterChanged = {};
+        var name1 = "paramName1";
+        var name2 = "paramName2";
+        var parameterChangedSpy1 = jasmine.createSpy("ParameterChangedSpy1");
+        var parameterChangedSpy2 = jasmine.createSpy("ParameterChangedSpy2");
+        panel.onParameterChanged = {};
+        panel.onParameterChanged[name1] = parameterChangedSpy1;
+        panel.onParameterChanged[name2] = parameterChangedSpy2;
+        panel.parameterChanged(param, name1);
+        expect(parameterChangedSpy1).toHaveBeenCalled();
+        panel.parameterChanged(param, name2);
+        expect(parameterChangedSpy2).toHaveBeenCalled();
+      });
+
       it("getParameterDefinition", function() {
         var promptPanel = {};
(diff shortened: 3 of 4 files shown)


plagoa:
@diogofscmariano @pamval @krivera-pentaho @annapopova1 @afrjorge @jvelasques please review.

Should we support the possibility of passing an array of parameter names to the EventAPI#parameterChanged function instead of a single parameter name? This is not specified on the issue, therefore I didn't go that way.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -339,7 +339,7 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
         dashboard: undefined,
 
         parametersChanged: false,
-        onParameterChanged: null,
+        onParameterChanged: {},
```
annapopova1: I prefer the previous initial value as null because is we do not register callbacks, will be checked only first if statement in 'parameterChanged' method.
Also IMO need add property description (js doc) for 'onParameterChanged' as for other class properties such as 'onBeforeRender'.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -641,8 +641,13 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          * @param {Object} value
          */
         parameterChanged: function (param, name, value) {
-          if (this.onParameterChanged) {
+          if (this.onParameterChanged && typeof this.onParameterChanged === 'function') {
             this.onParameterChanged(name, value);
+          } else {
+            var paramCallback = this.onParameterChanged[name];
```
annapopova1: We can have an exception in this line for situation: api.event.parameterChanged(null);

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -79,13 +79,25 @@ define([], function() {
...
      * @param {Function} callback   The function to be executed when the event is triggered.
      *                              Pass null if you wish to unbind this event
+     * @param {String} paramName    The name of the parameter to which the callback will be called
+     *                              This argument is optional. If no parameter name is passed on
+     *                              the callback will be assigned to all the parameters.
      * @example
      *  api.event.parameterChanged(function(parameterName, parameterValue) {
      *    // Execute event based code
      *  })
      */
```
annapopova1: Also we can add one more example (doc ) for registering callback for concrete parameter name.

plagoa:
@annapopova1 thanks for your comments, I'll update the PR accordingly.

plagoa:
Updated with Anna's comments.

@diogofscmariano @pamval @krivera-pentaho @annapopova1 @afrjorge @jvelasques please review.

annapopova1:
@plagoa I do not see how we can unregister current callback (set null value). You can check better this case if you will write unit test for it.

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -79,13 +79,35 @@ define([], function() {
...
+     *  api.event.parameterChanged(function(parameterName, parameterValue) {
+     *    // Execute event based code
+     *  }, "parameterName")
      */
-    this.parameterChanged = function(callback) {
-      api.operation._getPromptPanel().onParameterChanged = callback;
+    this.parameterChanged = function(callback, paramName) {
+      if (typeof callback === 'function') {
+        onParameterChanged = api.operation._getPromptPanel().onParameterChanged;
+        if (!paramName || paramName == '') {
```
annapopova1: In this statement enough using only first if (!paramName) because it also checks situation when paramName is empty.

plagoa:
@annapopova1 made that change. Please review.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -642,7 +645,18 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
...
-            this.onParameterChanged(name, value);
+            if (typeof this.onParameterChanged === 'function') {
+              this.onParameterChanged(name, value);
+            } else {
+              var paramCallback = this.onParameterChanged[name];
+              if (paramCallback) {
+                if (typeof paramCallback === 'function') {
+                  paramCallback(name, value); 
+                } else {
+                  console.log("The callback should be a function");
```
krivera-pentaho: This message doesn't tell me much of what is going on. Maybe something like "The parameterChanged callback for 'name' is not a function"

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -79,13 +79,35 @@ define([], function() {
...
+     * @example <caption>Register a parameter changed event to the parameter called "parameterName"</caption>
+     *  api.event.parameterChanged(function(parameterName, parameterValue) {
+     *    // Execute event based code
+     *  }, "parameterName")
      */
-    this.parameterChanged = function(callback) {
-      api.operation._getPromptPanel().onParameterChanged = callback;
+    this.parameterChanged = function(callback, paramName) {
+      if (typeof callback === 'function') {
+        onParameterChanged = api.operation._getPromptPanel().onParameterChanged;
```
krivera-pentaho: Is this supposed to be a global variable?

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -79,13 +79,35 @@ define([], function() {
...
+          onParameterChanged = callback;
+        } else {
+          if (!onParameterChanged) {
+            onParameterChanged = {};
+          }
+          onParameterChanged[paramName] = callback;
+        }
+        api.operation._getPromptPanel().onParameterChanged = onParameterChanged;
+      } else {
+        api.log.error("callback should be a function");
```
krivera-pentaho: Same comment for the message.

plagoa:
Updated with Kris' comments. 

@diogofscmariano @pamval @krivera-pentaho @annapopova1 @afrjorge @jvelasques please review.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -744,7 +747,18 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
...
-            this.onParameterChanged(name, value);
+            if (typeof this.onParameterChanged === 'function') {
+              this.onParameterChanged(name, value);
+            } else {
+              var paramCallback = this.onParameterChanged[name];
+              if (paramCallback) {
+                if (typeof paramCallback === 'function') {
+                  paramCallback(name, value); 
+                } else {
+                  console.log("The parameterChanged callback for '" + name + "' is not a function");
```
nantunes: Shouldn't this instead throw an error? And have a test for that behaviour.

(I was just passing by... nevermind if it doesn't make sense)

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -79,13 +79,31 @@ define([], function() {
...
+      var onParameterChanged = api.operation._getPromptPanel().onParameterChanged;
+      if (!paramName) {
+        onParameterChanged = callback;
+      } else {
+        if (!onParameterChanged) {
+          onParameterChanged = {};
+        }
+        onParameterChanged[paramName] = callback;
+      }
+      api.operation._getPromptPanel().onParameterChanged = onParameterChanged;
```
afrjorge: I think the local variable optimisation here is not working, e.g. if paramName is a falsy the code executes 3 writes (lines 97, 99 and 106) isn't it more efficient to just do the three write cases as such:
```
if (!paramName) {
  api.operation._getPromptPanel().onParameterChanged = callback;
} else {
  if (!api.operation._getPromptPanel().onParameterChanged) {
    api.operation._getPromptPanel().onParameterChanged = {[paramName]: callback};
  } else {
    api.operation._getPromptPanel().onParameterChanged[paramName] = callback;
  }
}
```
or if we cannot use ES6 computed property names:
```
if (!paramName) {
  api.operation._getPromptPanel().onParameterChanged = callback;
} else {
  var onParameterChanged = api.operation._getPromptPanel().onParameterChanged;
  if (!onParameterChanged) {
    onParameterChanged = {};
  }
  onParameterChanged[paramName] = callback;
  api.operation._getPromptPanel().onParameterChanged = onParameterChanged;
}
```
Also, if we execute the first example in the documentation and then execute the second example (we also want to execute a more specific callback for a parameter "A" changed event). If we do this in the inverse order (first execute example 2 and then execute example 1) do we want to lose the (parameter-specific) callback assigned in example 2? e.g.:
```
it("should allow to assign an overall callback and also a callback for a specific parameter", function() {
  var parameterCallback1 = function() {};
  var parameterCallback2 = function() {};
  eventApi.parameterChanged(parameterCallback1, 'param');
  eventApi.parameterChanged(parameterCallback2);
  expect(promptPanelSpy.onParameterChanged).toBeDefined();
  expect(promptPanelSpy.onParameterChanged['param']).toBe(parameterCallback1);
  expect(promptPanelSpy.onParameterChanged).toBe(parameterCallback2);
});
```
although this works (```eventApi.parameterChanged(parameterCallback2);``` executes first):
```
it("should allow to assign an overall callback and also a callback for a specific parameter", function() {
  var parameterCallback1 = function() {};
  var parameterCallback2 = function() {};
  eventApi.parameterChanged(parameterCallback2);
  eventApi.parameterChanged(parameterCallback1, 'param');
  expect(promptPanelSpy.onParameterChanged).toBeDefined();
  expect(promptPanelSpy.onParameterChanged['param']).toBe(parameterCallback1);
  expect(promptPanelSpy.onParameterChanged).toBe(parameterCallback2);
});
```

plagoa:
Made some changes to allow for the possibility of having callbacks definined for specific parameters and at the same time have a generic callback for the remaining parameters.

@diogofscmariano @pamval @krivera-pentaho @annapopova1 @afrjorge @jvelasques please review.

plagoa:
Updated PR to use Logger.warn instead of console.log.

krivera-pentaho:
This looks good to me

PR #573, opened 2016-01-25
[BACKLOG-5909] -- Added stateChanged event to the PromptPanel API

krivera-pentaho:


krivera-pentaho:
@plagoa @diogofscmariano @annapopova1 @jvelasques @pamval please review

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -446,6 +446,7 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
         onAfterRender: null,
         onBeforeUpdate: null,
         onAfterUpdate: null,
+        onStateChanged: null,
```
annapopova1: Need add property description (js doc) for 'onStateChanged' as for other class properties such as 'onBeforeRender'.

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -120,5 +120,21 @@ define([], function() {
         api.operation._getPromptPanel().ready = function(){};
       }
     };
+
+    /**
+     * Registers a listener for when the state of the prompt panel or parameter definition changes.
```
annapopova1: Probably for users will be useful to describe all state flags which changes affect to stateChanged event. What do you think?

package-res/resources/web/test/prompting/PromptPanelSpec.js
```diff
@@ -1363,6 +1363,24 @@ define([ 'dojo/number', 'dojo/i18n', 'common-ui/prompting/PromptPanel',
...
+
+          expect(panel.paramDefn).toBe(paramDefn);
+          expect(panel.onStateChanged).toHaveBeenCalledWith("promptNeeded", undefined, paramDefn.promptNeeded);
+          expect(panel.onStateChanged).toHaveBeenCalledWith("paginate", undefined, paramDefn.paginate);
+          expect(panel.onStateChanged).toHaveBeenCalledWith("totalPages", undefined, paramDefn.totalPages);
+          expect(panel.onStateChanged).toHaveBeenCalledWith("showParameterUI", undefined,paramDefn.showParameterUI());
+          expect(panel.onStateChanged).toHaveBeenCalledWith("allowAutoSubmit", undefined, paramDefn.allowAutoSubmit());
+          expect(panel.onStateChanged).toHaveBeenCalledWith("page", undefined, paramDefn.page);
+        });
+      });
```
annapopova1: In addition better add one more test to check that onStateChanged callback is not called.

package-res/resources/web/prompting/api/EventAPI.js
```diff
@@ -120,5 +120,21 @@ define([], function() {
...
+     * @method
+     * @param {Function} callback   The function to be executed when the event is triggered.
+     *                              Pass null if you wish to unbind this event.
+     * @example
+     *  api.event.stateChanged(function(name, oldValue, newValue) {
+     *    // Execute event based code
+     *  });
+     */
+    this.stateChanged = function(callback) {
+      api.operation._getPromptPanel().onStateChanged = callback;
```
annapopova1: I would also validate that the callback is really function before assign it to onStateChanged

annapopova1:
@krivera-pentaho What do you think about implemented PromptPanel#setState? Probably I should change it little bit to use method PromptPanel#setParamDefn that correctly call stateChanged callback.

annapopova1:
One more comment about "autoSubmit" and "allowAutoSubmit" flags.
I read description BACKLOG-5909 and BACKLOG-6092 more attentively. Thomas separated those flags. I think the "allowAutoSubmit" flag matches to ParameterDefinition#allowAutoSubmit. And the "autoSubmit" flag matches to "autoSubmit" property from PromptPanel. And I think we need call callback for "autoSubmit" flag also. And accordingly I will change api.operation.state.
What do you think about it?

package-res/resources/web/prompting/api/EventAPI.js
krivera-pentaho: Definitely, I will add them

package-res/resources/web/prompting/api/EventAPI.js
krivera-pentaho: Not that i don't agree, but we are not doing it in any of the other calls, so we should do it for all.

krivera-pentaho:
@annapopova1 if you use PromptPanel#setParamDefn then the onStateChanged would be called anyway. I don't think we need to explicitly call it in the PromptPanel#setState function, but it wouldn't hurt either.

Also, autoSubmit on the PromptPanel simply uses the "allowAutoSubmit" function value from the parameter definition. Is there somewhere else it is being set where I can trigger an onStateChanged call? It seems the only place is in the PromptPanel#setParamDefn call.

krivera-pentaho:
@annapopova1 @plagoa @jvelasques @diogofscmariano @pamval please review

annapopova1:
@krivera-pentaho allowAutoSubmit is initialized from server response and is set to paramDefn. PromptPanel#autoSubmit firstly is initialized from paramDefn but after that can be modified by submit component.

krivera-pentaho:
@annapopova1 thank you for the clarification

krivera-pentaho:
@annapopova1 @diogofscmariano @pamval @plagoa @jvelasques please review

annapopova1:
1. What about "parametersChanged" property? According to BACKLOG-6092 api.operation.state allows modify this property. But "onStateChanged" works for this property only into PromptPanel#parameterChanged. So if we use api.operation.state to modify "parametersChanged" property, we do not trigger "onStateChanged" for it. Is it correct?

2. I do not sure about PromptPanel#setParamDefn because it each time overwrite PromptPanel#autoSubmit property by property from paramDefn. Before using setParamDefn method, PromptPanel#autoSubmit property is set single time in constructor. What do you think about it?

krivera-pentaho:
@annapopova1 
1. I agree with this. I merged your PR so I will change that line to use setAutoSubmit
2. We definitely want to trigger onStateChanged when we use the state api call.
3. Do we want to change the prompt panel's property for being able to auto submit if we change the parameter definition, or should it only be changed the first time the parameter definition is set on the prompt panel?

diogofscmariano:
@krivera-pentaho 
3. I think the property should not be changed in the first time :)
Apart from this agree

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -531,9 +533,47 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
...
+            fireStateChanged("page", prevParamDefn, this.paramDefn, function(paramDefn) { return paramDefn.page; });
+          }
+        },
+
+        /**
+         * Sets the autoSubmit property on the PromptPanel
+         *
+         * @param autoSubmit {Boolean} The autoSubmit boolean
+         */
+        setAutoSubmit: function(autoSubmit) {
```
afrjorge: missing unit test

PR #579, opened 2016-01-27
[BACKLOG-5845] Prompt API - Implement submit

afrjorge:


afrjorge:
@annapopova1 @diogofscmariano @jvelasques @krivera-pentaho @pamval @plagoa  please review

package-res/resources/web/test/prompting/PromptPanelSpec.js
```diff
@@ -319,8 +319,10 @@ define([ 'dojo/number', 'dojo/i18n', 'common-ui/prompting/PromptPanel',
       });
 
       it("submit", function() {
+        spyOn(panel, "onSubmit");
         panel.submit();
         expect(panel.forceAutoSubmit).toBeFalsy();
+        expect(panel.onSubmit).toHaveBeenCalled();
       });
```
annapopova1: Probably better to add one more test to check the case when panel.onSubmit is not called

annapopova1:
good for me (without comment about tests)

package-res/resources/web/test/prompting/PromptPanelSpec.js
afrjorge: I don't think I will be able to test if PromptPanel#onSubmit wasn't executed when it's null (I get the message "Error: Expected a spy, but got null."). What I will do is improve this test by adding the parameters **promptPanel** and **options** that should be passed on to the onSubmit function.

annapopova1: For example,try to set some object (not function) and onSubmit should not be executed but should appear warning message.

afrjorge:
updated according to Anna's comments

@annapopova1 @diogofscmariano @jvelasques @krivera-pentaho @pamval @plagoa please review

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -677,7 +679,8 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          *
          * @name PromptPanel#_submit
          * @method
-         * @param {Object} options
+         * @param {Object}  [options]        Aditional configuration options.
+         * @param {Boolean} [options.isInit] Flag indicating if submit is being executed during initialization.
```
krivera-pentaho: Is this the only option? If so, then just make it a parameter, instead of wrapped in a JSON

afrjorge: A couple of times an object is provided with a property isInit. In the [PromptPanel](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L1289) and [ReportViewer](https://github.com/pentaho/pentaho-platform-plugin-reporting/blob/master/package-res/reportviewer/reportviewer.js#L670). [PIR](https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/package-res/resources/web/pir/view.js#L5369) doesn't provide the options parameter so I documented it as optional. I'm not sure if it's always the only property.

afrjorge:
updated according to Kris's feedback

@annapopova1 @diogofscmariano @jvelasques @krivera-pentaho @pamval @plagoa please review

PR #582, opened 2016-01-27
[BACKLOG-6049] boolean, function, object, and string specs

scottyaslan:
@dcleao @carlosrusso

### DIFF
--- a/package-res/resources/web/test/karma/unit/pentaho/type/boolean.Spec.js
+++ b/package-res/resources/web/test/karma/unit/pentaho/type/boolean.Spec.js
@@ -1,0 +1,81 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define([
+    "pentaho/type/boolean",
+    "pentaho/type/Context",
+    "pentaho/i18n!/pentaho/type/i18n/types"
+], function(booleanFactory, Context, bundle) {
+
+    "use strict";
+
+    /*global describe:true, it:true, expect:true, beforeEach:true*/
+
+    describe("pentaho/type/boolean -", function() {
+        it("is a function", function() {
+            expect(typeof booleanFactory).toBe("function");
+        });
+
+        describe("new boolean()", function() {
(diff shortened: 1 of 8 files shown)


package-res/resources/web/test/karma/unit/pentaho/type/boolean.Spec.js
```diff
@@ -0,0 +1,69 @@
...
+                PentahoBoolean = booleanFactory(new Context());
+            });
+            it("should return an object", function () {
+                expect(typeof new PentahoBoolean(1)).toBe("object");
+            });
+            it("should return a function", function () {
+                expect(typeof PentahoBoolean).toBe("function");
+            });
+            it("should accept 1 as true", function () {
+                expect(new PentahoBoolean(1)._value).toBe(true);
```
carlosrusso: please use the getter interfaces directly and the valueOf method:

expect(new PentahoBoolean(1).value).toBe(true); //getter
expect(new PentahoBoolean(1).v).toBe(true); // shorter version of .value getter
expect(new PentahoBoolean(1).valueOf()).toBe(true);

package-res/resources/web/test/karma/unit/pentaho/type/boolean.Spec.js
```diff
@@ -0,0 +1,69 @@
...
+                expect(function () {
+                    new PentahoBoolean(null)
+                }).toThrowError(bundle.structured.errors.value.isNull);
+            });
+            it("should not accept undefined", function () {
+                expect(function () {
+                    new PentahoBoolean(undefined)
+                }).toThrowError(bundle.structured.errors.value.isNull);
+            });
+        });
```
carlosrusso: Personally, I would also add some "sad path" cases:

expect(new PentahoBoolean("potatoes").value).toBe(false); //or .toBeNull(), whatever applies
expect(new PentahoBoolean(new Date()).value).toBe(false);

package-res/resources/web/test/karma/unit/pentaho/type/string.Spec.js
```diff
@@ -0,0 +1,69 @@
...
+                expect(function () {
+                    new PentahoString(null)
+                }).toThrowError(bundle.structured.errors.value.isNull);
+            });
+            it("should not accept undefined", function () {
+                expect(function () {
+                    new PentahoString(undefined)
+                }).toThrowError(bundle.structured.errors.value.isNull);
+            });
+        });
```
carlosrusso: expect(new PentahoString({}).value).toBe ?
expect(new PentahoString([]).value).toBe ?

package-res/resources/web/test/karma/unit/pentaho/type/boolean.Spec.js
dcleao: Hi, the `v` property is also an internal detail used for de-serialization/configuration purposes only. Please use the `value` property in this context.
`valueOf()` will be tested in the `Simple` tests.

package-res/resources/web/test/karma/unit/pentaho/type/string.Spec.js
dcleao: `PentahoString({}).value` should throw cause it is taken as a spec and the value is not there, thus it is `null` :-)

PR #589, opened 2016-02-01
[BACKLOG-6051] - Have validate behavior for foundation classes

dcleao:
### 1st commit
* Refinement types
    * Implementation, docs and unit tests
    * Discrete Domain Facet
* Validation changes
    * Moved validate from `Item` to `Value` again
        * `Property` does have a validate method, however it is of a different signature,
          so no advantage in having validate at the Item level
    * `Value#validate` now delegates calls to a new method: `Value.Meta#_validate(value)`
      * This way, validation behavior is defined by the meta type and can thus be
        configured using a meta-extends-spec, as is the case with inline types.
      * This also aligns better with refinement types in which the instance
        interface is meaningless; concentrating behavior in the meta type supports
        all cases: the meta type always exists.
    * `Value.Meta#validate(value)` is still the null/type safe version
    * The new `Value.Meta#validateInstance(value)` method can be used when value
      is knwon to be an instance of the type
    * Changed `Value.Meta#validate(value)` to return an error when given a `Nully` value
    * Introduced `pentaho/type/valueHelper` to hold two methods used in
      validate methods, `normalizeErrors` and `combineErrors`
* Docs
    * Improved documentation for validate* methods
    * Fixed jsdocs declared type for `Item.Meta#{view,viewClass}`
    * Removed all doclet @override tags cause it's causing doclets to be ignored...
    * Added missing doclet for the List.Meta class
* Fixes
    * Fix `Context~collectTypeIdsRecursive` not looking at the 'of' attribute
    * Fixed bug in `Base.js` where the instance returned by the constructor/init
      function would not be returned.
* Various
    * Added unit tests for Object#key
    * Added stub for `Item.Meta#inheritedStyleClasses`
    * Small change to `Item.Meta#to`: calling `this.context.create`, instead of `this.create`.
    * `Context#create` now also tests whether the given argument is already of the
       specified `baseType`.
    * Formatting some unit test files (4 to 2 tabs, missing ; and test naming)

### 2nd commit
* Removed Item.Meta.Mesa property
* Removed as many references to the word Mesa as I could, replacing
  by the more familiar "instance constructor" expression, whenever possible.
* Removed Item.Meta#proto as it wasn't really being used
* Moved the Item.Meta description to Item and removed the Item description.
  * Reworded the Item.Meta description focusing on the dual of
    "instance" and "type metadata" classes, keeping as much as possible
    from Carlos' description.
* Fixed some tests that the previous commit broke... (Item.Meta#{is,to})
* Fixed missing facets namespace in docs

@graimundo, @scottyaslan, @davidmsantos90, @nantunes, @marcovala, @carlosrusso please provide your valuable feedback.


nantunes:
Ok. Changes understood conceptually. Units tests validated and passing. @graimundo should merge when possible, so we can base the remaining work on this code base.

PR #591, opened 2016-02-01
[BACKLOG-6051] Add unit tests and dev doc changes

nantunes:
Removed Property.validation.Spec.js, it only makes sense in the context of a Complex (some tests were duplicated).

Removed the possibility of validate returning a single Error instance (it is always a non-empty array of `Error` or `null`).

@dcleao please validate.


nantunes:
Invalidated by separate development.

PR #594, opened 2016-02-01
[BACKLOG-6049] - Added tests for valueHelper

graimundo:
- Complemented Value unit tests


graimundo:
@dcleao please review

PR #597, opened 2016-02-01
[BACKLOG-6051] Allow overriding a property with a refinement

nantunes:
@dcleao please review. Needed for sandbox's calc demo.


dcleao:
Good catch. Need to restrict the condition you added a bit more, but I totally missed that class of tests :-(

nantunes:
Thank the sandbox. It is proving its value.

My condition is _naive_. Please further elaborate it.

PR #603, opened 2016-02-02
[BACKLOG-6049] Create missing unit tests and update documentation for simple.js

davidmsantos90:
@dcleao @graimundo please review


graimundo:
@davidmsantos90 
Missing API reference documentation for set value.
valueOf method also seems like another alias for get value.

PR #608, opened 2016-02-02
[BACKLOG-6249] DEV: Create missing unit tests and update documentatio…

dcleao:
…n for Context.js

@graimundo please review.

### DIFF
--- a/package-res/resources/web/pentaho/type/Context.js
+++ b/package-res/resources/web/pentaho/type/Context.js
@@ -15,6 +15,7 @@
  */
 define([
   "module",
+  "./Item",
   "../i18n!types",
   "./standard",
   "../lang/Base",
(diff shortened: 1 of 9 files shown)


graimundo:
@dcleao missing tests for collectTypeIds and collectTypeIdsRecursive.
There are also some branching conditions which are not covered by the Context.Spec (from Context.js lines 137, 203, 209, 246, 267, 286, 294, 309 and 322).

graimundo:
@dcleao disregard my previously deleted comment. Unit tests and coverage are great.
My only question is if there isn't anything that could be documented on those private _get* methods that would aid in their understanding? 
Besides that I would say go with the merge.

dcleao:
Merging it myself. I'll add additional docs for private stuff at a later time.

PR #610, opened 2016-02-03
[PPP-3455] - Dynamic code injection vulnerabilities found

akhayrutdinov:
- replace `eval()` with `JSON.parse()` in models-svc.js, ref.js

@rfellows, @krivera-pentaho, @diogofscmariano, review it please

### DIFF
--- a/package-res/resources/web/dataapi/models-svc.js
+++ b/package-res/resources/web/dataapi/models-svc.js
@@ -18,7 +18,7 @@
 /*
                          pentaho.pda.SvcHandler
 */
-define("common-data/models-svc", ['common-data/oop', 'common-data/controller'], function(){
+define("common-data/models-svc", ['common-data/oop', 'common-data/controller', 'dojo/request'], function(oop, controller, request){
 pentaho.pda.SvcHandler = function svcHandler(sandbox) {
 	pentaho.pda.Handler.call(this, sandbox);
     this.type = pentaho.pda.SOURCE_TYPE_SVC;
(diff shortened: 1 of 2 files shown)


package-res/resources/web/dataapi/models-svc.js
```diff
@@ -55,7 +55,7 @@ pentaho.pda.SvcHandler.prototype.getSources = function(callback, options) {
             var contextName = options['context'];
             var query = "domainName=&"+((contextName)? 'context='+contextName : "");
             var jsonStr = pentahoPost( url, query, null, 'text/text' );
```
rfellows: I would prefer that we move away from pentahoPost and to the dojo request method. you can specify the return type and not have to parse it after the fact.

akhayrutdinov: @rfellows, I'm afraid it is a bit wider, than expected from this PR. That is not a problem for me to introduce such changes, but it misleads from the focus of the ticket.

@pamval, @mbatchelor, do you agree with Rob (replacement of `pentahoPost`)

pamval: I'd go ahead with that change here, in this call in this file. 
Not a global replacement of pentahoPost everywhere.

krivera-pentaho: Would we want to consider replacing the implementation of pentahoPost to use the dojo request method?

rfellows: no, there are probably tons of uses of it out there still and we'd have to make it return exactly what it returns now, which is potentially still using eval, so it wouldn't really provide the whole solution without refactoring all of the places that use it as well.

akhayrutdinov:
@rfellows, I've updated the PR with using `dojo/request`. Can you please review it? I haven't found any examples of using `request.post()` in the project, thus I'm not sure I migrated from the old API correctly

package-res/resources/web/dataapi/models-svc.js
```diff
@@ -54,8 +54,20 @@ pentaho.pda.SvcHandler.prototype.getSources = function(callback, options) {
             var url = this.SERVICE_URL+'?action=listmodels';
             var contextName = options['context'];
             var query = "domainName=&"+((contextName)? 'context='+contextName : "");
-            var jsonStr = pentahoPost( url, query, null, 'text/text' );
-            var rawModels = eval( '('+jsonStr+')' );
+
+            var rawModels;
+            request.post(url, {
```
rfellows: If we aren't sending anything in the data portion of the request, why bother with post? wouldn't a get suffice?

akhayrutdinov: @rfellows, I wish I knew the answer :smile: 
That is how is was implemented, I'm surprised with POST too

rfellows: ok, i'll merge it then

PR #611, opened 2016-02-03
[BACKLOG-6049] Added unit tests to PropertyMetaCollection.

carlosrusso:
@dcleao @scottyaslan please review


nantunes:
@graimundo all ok.

But with four additional tests it would reach 100% coverage (93.33% as is).

dcleao:
looks good :-)

PR #613, opened 2016-02-04
[DOC] Upgrade AMD tag for common-ui use cases and updated the documentation

davidmsantos90:
@dcleao @graimundo please review

This is the output:

amd-tag

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ eclipse-bin/
 doc-js
 build-res/pentaho-js-build/
 build-res/module-scripts/
+build-res/pentaho-js-build.zip
 build-res/requireCfg-raw.js
 build-res/requireCfg.js
 test-res/data/sampledata.log
--- a/build-res/jsdoc-template/static/styles/jsdoc-overrides.css
+++ b/build-res/jsdoc-template/static/styles/jsdoc-overrides.css
@@ -1,11 +1,18 @@
-
 h3 {
   font-size: 22px;
 }
 
-.amd-module {
-	font-size: 120%;
+#main {
+  overflow-x: hidden;
+}
+
+.amd-type, .amd-module h5 {
+  font-size: 120%;
   line-height: 140%;
   margin-bottom: 1em;
   margin-top: 1em
+}
+
+.amd-module pre {
+  min-width: 700px;
 }
--- a/build-res/jsdoc-template/tmpl/amd.tmpl
+++ b/build-res/jsdoc-template/tmpl/amd.tmpl
@@ -2,8 +2,30 @@
 var data = obj;
 var self = this;
 ?>
-<?js if(data.amd != null) { ?>
-	<div class="amd-tag">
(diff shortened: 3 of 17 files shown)


dcleao:
@diogofscmariano

dcleao:
I like this. I think it allows for a more concise solution than using the `@amd` tag in member doclets for "singleton classes" and "factory classes".

For "singleton classes", would this solve the issue?
```js
@name pentaho.type.ValueHelper
@class
@amd {pentaho.type.ValueHelper} pentaho/type/valueHelper
```

Note that adding the type in the `@amd` tag changes the return type of the module from: the constructor, `Class.<pentaho.type.ValueHelper>`, to an instance of `pentaho.type.ValueHelper`.

Not sure that important subtlety is visible enough!

---
My concern is that this raises another issue — that of the required argument name in the code snippet. Being a factory, the default "use the name of the class", `Boolean`, does not work well anymore — the argument should be called something like `booleanFactory`.

1. Support another sibling tag, like `@amd-argument booleanFactory` where one could specify the name of the argument
2. Allow some way to specify the whole `require(["..."], function(booleanFactory) {});`
3. Use the last segment of the amd module id to name the argument: would result in `function(boolean) {}` — which is better as only classes/constructors should be uppercase.

I fear 2) would result into:
* people getting out of the norm in the parts of the statement that should be normalized, like `/*code goes here*/` and such
* adding full examples in there

So I'd go with option 1), or 3).

Option 3) is the simplest. And can serve as default, even if option 1) is chosen.
Option 1) might not be that hard: add a new, subordinate, tag that the main `@amd` tag would sneak into when rendered.

dcleao:
This makes me realize that we could document the normal class case with a type annotation of `Class.<pentaho.type.ValueHelper>`. Because the idea, I thing, is that when unspecified we'd show nothing.
This way, we'd always output the type to make things clear.

graimundo:
I say go with option 3). That with the already existing return type should be enough to clarify any ambiguities.

diogofscmariano:
```
@name pentaho.type.ValueHelper
@class
```
This is not good. The name tends to override many things...

On cdf i'm using the last segment. I would just like to have that better defined :)
Just another question... we are doing this in 2 different places, shouldn't we have a repo for this and stop doing 2 different commits every time we change the template? @pamval?

dcleao:
@diogofscmariano I didn't understand what it is that you are saying isn't good.

scottyaslan:
Updated…please take a final look.

Thanks,
-Scotty

diogofscmariano:
Using the name tends to do override other parts of the documentation. From jsdoc:

> Warning: By using the @name tag, you are telling JSDoc to ignore the surrounding code and treat your documentation comment in isolation. In many cases, it is best to use the @alias tag instead, which changes a symbol's name in the documentation but preserves other information about the symbol.

dcleao:
Ah, I see what you mean. The doclet was just an example, with no surrounding code — a virtual doclet.
I thought you were somehow referring to the matter under discussion.

dcleao:
@graimundo I guess this one is waiting for your additional comments or merge :-)

graimundo:
Closing this PR as it was decided that for the meantime the JSDoc template will be hosted in the CDF repository. This effort will be merged there until we have a central repository from where to service the JSDoc template.

PR #614, opened 2016-02-04
[BACKLOG-6249] DEV: Create missing unit tests and update documentation for Context.js

dcleao:
* Added lots of documentation
* Moved code of Context#create to the already existing Value.Meta#create
* Removed Context#get support for unlikely use cases: by Value.Meta and by Value instance.
  These now throw a dedicated error instead of passing unnoticed or throwing a
  cryptic error

@graimundo, @scottyaslan, @davidmsantos90, @nantunes, @marcovala, @carlosrusso please provide your valuable feedback.


package-res/resources/web/pentaho/type/Context.js
```diff
@@ -57,24 +57,39 @@ define([
...
+   * Context variables are specified at construction time, through same named keyword arguments.
+   *
+   * These variables are used to select applicable configuration rules
+   * that are applied to the types provided by the context.
+   *
+   * Because configuration rules rely on the identification of types,
+   * only non-anonymous types -
+   * those that have an [id]{@link pentaho.type.Value.Meta#id} -
+   * can be _directly_ configured.
+   *
```
carlosrusso: What about adding a paragraph that explains why we need a context?

It would help visualization developers understand that they should invoke the Value classes from the context instead of accessing them directly.

dcleao: This kind of says why:
«These variables are used to select applicable configuration rules that are applied to the types provided by the context.»
This is, the big reason is _configuration_. But I will try to give more emphasis to that :-)

carlosrusso: Yes, I noticed, but the level here is a bit too deep. I felt that a less technical and more motivational explanation was needed in order to capture the interest for this core concept and to give an idea of its systemic influence.
Mind you, I know it's hard to come up with such an explanation, I've been trying to come up with one myself to better illustrate my point...

package-res/resources/web/pentaho/type/Context.js
```diff
@@ -155,15 +240,63 @@ define([
...
-     * Gets the instance constructor of a type.
+     * Gets, asynchronously, the **configured instance constructor** instance constructor of a type.
+     *
+     * For more information on the `typeRef` argument,
+     * please see [UTypeReference]{@link pentaho.type.spec.UTypeReference}.
+     *
+     * This method can be used even if a generic type metadata specification references
+     * non-standard types whose modules have not yet been loaded by the AMD module system.
+     *
+     * An error is thrown when:
```
carlosrusso: Do you disapprove the "@throws" tag? Or you simply thought it would be better to describe the thrown errors in the text?

package-res/resources/web/pentaho/type/Context.js
```diff
@@ -306,6 +390,34 @@ define([
       throw error.argInvalid("typeRef");
     },
 
+    /**
+     * Gets the instance constructor of a type given it's id.
```
carlosrusso: typo: it's -> its: given its id

package-res/resources/web/pentaho/type/value.js
```diff
@@ -231,6 +231,187 @@ define([
...
+         *   var context = new Context({container: "data-explorer-101"});
+         *   var Value   = context.get("value");
+         *
+         *   var productList = Value.meta.create({
+         *         _: [{
+         *           props: ["id", "name", {name: "price", type: "number"}]
+         *         }],
+         *
+         *         d: [
+         *           {id: "mpma", name: "Principia Mathematica", price: 1200},
```
carlosrusso: I thought you meant Newton's book "Philosophiæ Naturalis Principia Mathematica", which would have been priceless, but from the context it seems you're referencing Whitehead's and Russel's book.

package-res/resources/web/pentaho/type/_doc/spec/UTypeReference.jsdoc
```diff
@@ -0,0 +1,211 @@
+/*!
+ * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
```
carlosrusso: 2016

package-res/resources/web/pentaho/type/Context.js
dcleao: No I don't. I simply didn't use in the remainder of the code, and didn't feel like starting to use it now. I guess I am subconsciously relying on a subsequent pass through all the code to fix that.
An interesting case arises in this use, though: that rejections and throws, being similar conditions, would be documented differently, if we were to use the `@throws` clause.

package-res/resources/web/pentaho/type/value.js
dcleao: I am referring to that. I'm not referring to the other guys' "review". Do you mean the full title is not what I put there? I'll check that.

dcleao: Sorry, I was in fact referring to Russel's book on Mathematics.

package-res/resources/web/pentaho/type/Context.js
dcleao: Another thing comes to my mind. Document that the method throws an error when an argument documented as being "required" is passed might be an anti-pattern. There's so much about a documented contract that we don't validate with errors. We just check the most obvious, causes of misuse, to aid the programmer in the programming effort and to find coding errors quicker.
These would be the errors that a compiler would catch, by static analysis — requiredness and type conformance — most of these we don't test because it would be expensive in performance and verbose. We could do this if we had a DEBUG variable setup along with dead-code elimination (we can actually do this with uglify).

What I do think should be documented as thrown errors are things like invalid arguments and operations.

Now, independently of the documentation, whatever we code, we need to test. So, even if we don't document the fact that no specifying a required argument actually throws an error, we need to test it :-)

carlosrusso: To test violations of contract or not, that is the question.

I agree that adding code just to ensure a required argument is indeed passed to a function or constructor will put a performance penalty that might be avoided altogether if we assume the documentation is properly followed. That idea about DEBUG variable seems to be great, specially if is only used in a block like:

```js
if(DEBUG){
  // validate arguments
  if (...) throw ...
  if (...) throw ...
  if (...) throw ...
}
```

Personally, I find it currently hard to predict which cases fall into the category of "no need to test that"

dcleao: With all the thinking, I might have not been clear with the resulting desire:
```js
/**
 * Gets a _foo_ given a bar.
 * @param {object} bar A bar to get a foo from.
 * @return {object} A foo.
 * @throws {Error} Invalid argument. When `bar` is closed.
 */
function foo(bar) {
  if(bar == null) throw error.argRequired("bar");
  if(DEBUG) {
    if(bar.closed) throw error.argInvalid("bar", "Bar is closed. Cannot get foo.");
  }

  // ...
}
```

So:
1. Implement `argRequired` errors
2. Implement `argInvalidType` errors in some cases; preferably wrapping in DEBUG block
3. Don't document `argRequired` or `argInvalidType` errors (they're implied in the contract)
4. Document and implement `argInvalid`, `operInvalid` errors (without DEBUG block!)

What about the use of `@throws`?

dcleao:
@carlosrusso: updated the Context description. Let me know if this is better or worse :-P

package-res/resources/web/pentaho/type/Context.js
dcleao: Unfortunately, common-ui does not build JS at the moment, so supporting elimination of the DEBUG variable isn't for free.

carlosrusso: The analysis above seems reasonable, mostly because it adheres to the principle of a single-source-of-truth. The basic difference is that sometimes the truth is in the code, sometimes it is in the documentation.

The hybrid cases `argInvalid` and `operInvalid` are circunstancial and probably too complex  to be avoided by resorting to documentation only.

As for `@throws`, we should probably include them all because in a way it's one kind of output.
Devs invoking the code in a try-catch block would probably need to know which exceptions are there.
A block of `@throws` has the potential of becoming easily unreadable.

package-res/resources/web/pentaho/type/Context.js
```diff
@@ -56,25 +56,91 @@ define([
   /**
    * @name pentaho.type.Context
    * @class
-   * @implements pentaho.type.IContext
-   * @classDesc The `Context` class contains _Value_ instance classes
-   * _configured_ for a particular _context_.
+   * @classDesc A `Context` object holds instance constructors of **configured** _Value_ types.
+   *
```
carlosrusso: What about something along the lines:

>  When a visualization is being assembled it should not necessarily be unaware of the environment where it is going to be used. 
> A context object gathers information that has a global scope, such as the current locale or the current theme, which is likely to have an impact on how the visualization should be presented to the user. For instance, the default color palette used in a categorical bar chart might be related to the current theme.
> As such, the context object should contain the necessary logic to facilitate the configuration of the building blocks of the visualization.
> The Pentaho Component Metadata Model embraces this concept by defining most types as object factories that take a context as their argument.

package-res/resources/web/pentaho/type/Context.js
dcleao: Let me rewrite the aforementioned rules:
1. `argInvalidType` and `argRequired`
  * already documented in an argument's tag
  * sometimes these are actually implemented and thrown — but simply because it helps catch common programming bugs faster
  * a `@throws` clause would be redundant and would document a circumstantial extra-precaution being taken
  * if we implement the error, we need to test it as well (or coverage drops), but that doesn't mean we need to document it
  * when implemented, these are like **assertions**
  * if the code were compiled, most cases would be caught at compile time
  * if the code were compiled, most errors would never reach a try/catch
  * there is a gotcha in these two statements: to catch _nully_ at compile time, the compiler would need to perform "null analysis", something which is hard to perform
  * however, the main and relevant distinction is between good use and bad use of a contract — errors thrown in the course of good/acceptable/unforseeable use are documented and can/should be handled by callers, while errors thrown in the course of bad use (invalid program) are not.
  * this is akin to how in Java only some types of errors are required to be part of method signatures
  * last, but not the least, it would be overkill to document these errors; it would be lengthy, and it would actually feel redundant.
2. `argInvalid` and `operInvalid`
  * document using `@throws`
  * implement, using `throw`
  * these conditions depend on the flow of the program and the state of its objects and are utterly impossible to detect at compile time
  * mostly these will still be considered the result of a "bad program"
3. "timeout", "user validation", ...?
  * document using `@throws`
  * implement, using `throw`
  * these conditions generally depend on state and actions that are external to the program and whose occurrence it thus can't control or foresee, but only expect and prepare for

Questions that I still think need discussion is:
* how to handle the difference between rejections and throws; there's an extensive discussion on this subject in https://github.com/jsdoc3/jsdoc/issues/509
* should we start using `@throws` now, or do it all in a full replace do the codebase?

lucboudreau: The proper way to add sanity checks during development without impacting production is not the BEBUG variable, You should use the 'assert' feature built right into the JVM. These checks are enabled at test time, but usually turned off during prod with no overhead penalty.

https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html

dcleao: @lucboudreau this is JavaScript and we don't have the fortune of an assert statement :-(

lucboudreau: lol ok. Didn't see that. Will go grab more coffee..

package-res/resources/web/pentaho/type/Context.js
dcleao: Thanks! Added your text :-)

graimundo:
Looks good. @pentaho/millenniumfalcon let's take note on the rules of argument validation discussed here and use them from now on.

PR #615, opened 2016-02-04
[CLEANUP] - Upgrading PhantomJS to 2.1.3

dcleao:
* Removed `config/phantomjs-shims.js` file which is not needed anymore for this phantomjs version
* Took the opportunity to update most used packages in `package.json` to their latest versions

Finally, a PhantomJS version that is mostly fully ES5 compatible (http://kangax.github.io/compat-table/es5/)

@pentaho-nbaker, @pamval, @pentaho/millenniumfalcon, @pentaho/orlandocalisbon please review and comment on any impediments


pamval:
Shouldn't we update phantom.js across the board, on all projects  ?

dcleao:
@pamval I think you're right. Could common-ui serve as the experiment?

I think each project needs to be handled specially — a task in a story for each?

The packages that each use differ. Frequently, upgrading one package causes needing to upgrade others (peerDependencies).
Because test-related packages are upgraded, some tests may start breaking, for example, due to deprecated jasmine syntax.
Also, some things, like karma.conf.js may need to be changed  to accommodate for problems that occur on a specific regex pattern we usually use in the files config (e.g. [change](https://github.com/pentaho/pentaho-platform-plugin-common-ui/commit/a1cbbdeff819a6068be29a3ee97e4f82efbf38b7#diff-74a0d66dcdf89050c27b9d449fbfb1ceR15))

dcleao:
Rebased.

dcleao:
So the silence would mean that everyone is in agreement with merging this ;-) ?

CodeOnCoffee:
Lets see how it behaves and then propagate it around afterward

dcleao:
Success! Build succeeded installing the new PhantomJS.

PR #618, opened 2016-02-05
[CLEANUP] - Moving common-ui tests to test-js folder

dcleao:
#### Commit 1
* Moved `config/require-config.js` to `test-js/unit/require-config.js`.
  All other tool config files remain in the config folder.
* The file `test-js/unit/require-config.js` now only contains per-project,
  specific AMD configurations.
* The new file `config/context-end.js` now contains all of the code
  necessary to bootstrap the the tests and contains no project specific
  AMD configuration
* The old `config/context.js` was aptly renamed to `config/context-begin.js` and
  now contains two new handy dev-time configuration variables (check them out!)
* Merged one of the `models-mqlSpec.js` with the other same named,
  but in a different folder test file
* Moved some test files to match the source folders of the files being tested

#### Commit 2

I noticed that if only the prompting api tests were ran, these would fail, apparently, randomly. This had to do with some lost `jQuery` identifiers, and some undeclared `$` dependencies. Also, there's a strange test explicitly spying on `window.$` (accepting it as global?), that was failing.
So I propose these fixes. @pentaho/orlandocalisbon tell me if you agree with these. Note that the tests work if all common-ui tests are run.

@pentaho-nbaker, @pamval, @pentaho/millenniumfalcon, @pentaho/orlandocalisbon please review.


diogofscmariano:
Agree (or most of it). Thanks for tackling that down @dcleao. We did the changes on the code side, but tests were running probably due to other dependencies.
I just dont like the todos. @jvelasques can you please help here?

graimundo:
Looks good to me. When the spying on the `window.$`is tackled we should merge.
Btw, those dev config vars are indeed handy :+1:

test-js/unit/prompting/PromptPanelSpec.js
```diff
@@ -546,14 +546,20 @@ define([ 'dojo/number', 'dojo/i18n', 'common-ui/prompting/PromptPanel',
             }
           });
           comp.type = "ScrollingPromptPanelLayoutComponent";
-          spyOn(window, "$");
+
+          // TODO: Find alternative for:
+          //spyOn(window, "$");
```
jvelasques: @dcleao we could have something like:
`var _jqHolder={$:$};
spyOn(_jqHolder, "$")`
ant then the expect would be
`expect(_jqHolder.$).not.toHaveBeenCalled()`

dcleao: Ok. I just need to be sure it is the same jQuery instance that the source code is using. I say this because in some places it is common-ui/jquery-clean and in others cdf/lib/jquery

dcleao: I don't think that rewrite will work: you can intercept the holder property, placing a spy function instead of jquery, but that doesn't say a thing about the actual jquery function being called...

dcleao: So, for now, I'd accept this as is (commented with TODO), and later you guys can replace it with some alternative to testing what this is testing. What do you think?

dcleao:
Just wanna note that this PR has a short mergeable-status duration.
Any PR that gets accepted before and has tests (all have?) will be a big pain to conciliate on my side...

dcleao:
Hi guys, I removed the 2nd commit, and will initiate a separate PR with that alone. They are actually independent issues. Sorry for that.

dcleao:
@pentaho-nbaker, @pamval, @pentaho/millenniumfalcon, @pentaho/orlandocalisbon please review.

dcleao:
Hi ho! Just to remind you guys of this — expiry date should be close...
Sorry and thanks.
@pentaho-nbaker, @pamval, @pentaho/millenniumfalcon, @pentaho/orlandocalisbon please review.

graimundo:
As agreed with @diogofscmariano merging this. We will address the jquery spy on in a separate commit.

PR #620, opened 2016-02-08
[BACKLOG-6264] Tests and documentation of pentaho.visual.base

carlosrusso:


package-res/resources/web/common-ui-require-js-cfg.js
```diff
@@ -202,4 +202,10 @@
   // Sample visualizations
   requireService["pentaho/visual/samples/calc/visualTypeProvider"] = "IVisualTypeProvider";
 
+  requireCfg.packages = requireCfg.packages || [];
```
dcleao: Please remove this check as it is always defined.

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -22,89 +22,180 @@ define([
...
    * @classDesc This is the base class for visualizations.
    *
+   * A container is expected to instantiate a `View` with a reference to a `Model` instance,
+   * which may not be immediately valid.
+   *
+   * Over time, the container mutates the `Model` instance and triggers events.
+   * In response, the method [_onChange]{@link pentaho.visual.base.View#_onChange} is
+   * invoked to process the events, which may cause the `View` to update itself.
+   *
    * @description Creates a base `View` instance.
```
dcleao: This description is strange as the class is abstract. Maybe "Initializes a `View` instance." ?

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -22,89 +22,180 @@ define([
...
+   * An error is thrown if this is not a valid DOM element.
    * @param {pentaho.visual.base.Model} model The base visualization `Model`.
+   * @throws {Error} Argument required. The constructor must be invoked with two arguments.
+   * @throws {Error} Argument invalid. A valid DOM element must be passed to the constructor.
    */
 
   var View = Base.extend(/** @lends pentaho.visual.base.View# */{
     constructor: function(element, model) {
-      if(!isElement(element)) {
+      if(arguments.length === 0)
```
dcleao: This is original. However, I don't think this is a good idea. We should have a single error message for a single error condition and in this code we're, in practice, validating the requiredness of `model` argument twice. Also, we should try to make the cases `nully` and absent/missing to mean and _result_ the same, whenever possible.

test-js/unit/pentaho/type/number.Spec.js
```diff
@@ -17,45 +17,57 @@ define([
...
+      it("should accept a string '1' and return the number 1", function() {
         expect(new PentahoNumber('1').value).toBe(1);
       });
 
-      it("should throw", function () {
-        expect(function () {
-          new PentahoNumber('one').value;
-        }).toThrowError(bundle.format(bundle.structured.errors.value.cannotConvertToType, ['Number']));
+      it("should throw and not accept a 'non-numeric' argument", function() {
+        [
```
dcleao: Did you intend to add additional test values?

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,57 @@
...
+      var data = {};
+      dataSpec = {v: data};
+    });
+
+    describe("the constructor ", function() {
+
+      it("should throw if not invoked with two arguments", function() {
+        expect(function() {
+          return new View();
+        }).toThrowError(/required/);
```
dcleao: How do I know which of the required errors is being thrown? Why not follow the pattern we've been using? I know you dislike it, but unless you propose a better alternative, I'd prefer staying with what we've got.

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,57 @@
...
+
+        expect(function() {
+          return new View(element);
+        }).toThrowError(/required/);
```
dcleao: idem

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,57 @@
...
+      });
+
+      it("should throw if the first argument is not a DOM element", function() {
+        [
+          null, 1, true, {}, []
+        ].forEach(function(elem) {
+
+          expect(function() {
+            return new View(elem, dataSpec);
+          }).toThrowError(/Invalid/);
```
dcleao: idem

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,57 @@
...
+        expect(function() {
+          return new View();
+        }).toThrowError(/required/);
+
+        expect(function() {
+          return new View(element);
+        }).toThrowError(/required/);
+
+        expect(function() {
+          return new View(element, dataSpec);
```
dcleao: Sending in an arbitrary object into the `model` argument doesn't seem like good idea. It is true that you and I know that the current code does not validate the type of thing received, or uses it in the constructor somehow, but relying on that on a unit test looks like a brittle assumption and unnecessary risk.

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,57 @@
...
+          expect(function() {
+            return new View(elem, dataSpec);
+          }).toThrowError(/Invalid/);
+        });
+      });
+
+    });
+
+    it("should not be renderable (returns a rejected promise when render() is called)", function() {
+      var view = new View(element, dataSpec);
```
dcleao: Here we go. Why is this being rejected? Nor the test title nor its assertion say why. I suspect it is throwing because dataSpec does not have a `dataSpec.meta.validate method`? And not because it is invalid, as I would expect :-)

test-js/unit/pentaho/visual/base/modelFactory.Spec.js
```diff
@@ -0,0 +1,141 @@
...
+      }).not.toThrowError();
+    });
+
+    it("can be instantiated without arguments", function() {
+      expect(function() {
+        return new Model();
+      }).not.toThrowError();
+    });
+
+    it("can be instantiated even if the modelSpec is not an object ", function() {
```
dcleao: Clearly, I dislike this unit test. It's not something you wanna test and set in stone (for us and our api's users) whether certain arbitrary values that explicitly go against the documented contract do not throw and result in being equivalent to a `nully` argument. This is circumstantial behaviour that can change with any rewrite of the code. The values `null` and `undefined` are expected to be equivalent to not passing any arguments. All the others are (for me) unwanted, undesired assertions.

test-js/unit/pentaho/type/number.Spec.js
carlosrusso: Oops

test-js/unit/pentaho/visual/base/View.Spec.js
carlosrusso: This is not an attempt to boycott the pattern. I just messed up and failed to see the obvious.

test-js/unit/pentaho/visual/base/View.Spec.js
carlosrusso: Good point. The promise is rejected because ._render throws  "Not Implemented"

test-js/unit/pentaho/visual/base/modelFactory.Spec.js
carlosrusso: This is not the test you are looking for.

test-js/unit/pentaho/visual/base/View.Spec.js
carlosrusso: How else should we test if two arguments are being passed to the constructor? If I pass a nully it's as if I omitted the argument, if I pass a simple {}, I might be sending the wrong message.

The alternative of removing the check from the constructor does not please me.

package-res/resources/web/pentaho/visual/base/View.js
carlosrusso: Makes sense.

carlosrusso:
@dcleao @pentaho/millenniumfalcon  please review. 
Accurate feedback will be important to ensure good work quality, so please unleash the critic in you.
Now is the time.

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -22,89 +22,185 @@ define([
...
+      var validationErrors = this.model.validate();
+      return validationErrors;
+    },
+
+    /**
+     * Determines if the visualization is in a valid state.
+     *
+     * A visualization in an invalid state should not be rendered.
+     *
+     * @return {boolean} Returns `true` if this visualization is valid, or `false` if not.
```
swagner8412: Returns `true` if this visualization is valid, or `false` if not valid.

test-js/unit/pentaho/visual/base/View.Spec.js
dcleao: So I might not have been clear here. What I meant is that you should not pass something other than a model object into the second argument: new Model().

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,56 @@
...
+          return new View(element);
+        }).toThrowError("Argument required: 'model'.");
+      });
+
+      it("should throw if the first argument is not a DOM element", function() {
+        [
+          "div", 1, true, {}, []
+        ].forEach(function(elem) {
+          expect(function() {
+            return new View(elem, dataSpec);
```
dcleao: I hadn't seen your response to my comment on the initial commit. I've answered there now. Here's what I wrote there:

> So I might not have been clear here. What I meant is that you should not pass something other than a model object into the second argument: new Model().

This applies to all `dataSpec` values being passed to argument `model` in this file.

carlosrusso: Makes sense.

carlosrusso:
@dcleao take a fresh look, please

test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -0,0 +1,67 @@
...
+    var element, dataSpec, Model, model;
+
+    beforeEach(function() {
+      element = document.createElement("div");
+      dataSpec = {
+        width: 1,
+        height: 1,
+        data: {v: {}}
+      };
+      Model = modelFactory(new Context());
```
carlosrusso: @dcleao: Should I have used instead
```js
beforeEach(function(done) {
  ...
  var context = new Context();
  Model = context.getAsync(modelFactory).then(done);
});
```

dcleao: No, the new formulation looks good (besides, the sync/async hybrid formulation you're proposing does not really work, if I read it correctly :-P ). 
You don't need to use the `getAsync` method because the `get` method will succeed, because when given a type factory, like `modelFactory`, there's no need to load any other AMD modules.

carlosrusso:
@dcleao Please take another look.

dcleao:
Looks great to me :+1:

dcleao:
Anyone else?

PR #623, opened 2016-02-09
[BACKLOG-6293] DEV: Create unit tests and documentation for pentaho/util/promise.js

dcleao:

### DIFF
--- a/package-res/resources/web/pentaho/util/promise.js
+++ b/package-res/resources/web/pentaho/util/promise.js
@@ -14,27 +14,105 @@
  * limitations under the License.
  */
 define([
+  "./arg",
+  "./error",
   "../shim/es6-promise"
-], function() {
+], function(arg, error) {
   "use strict";
 
-  return {
-    call: function(fun, ctx) {
-      // Wrapping the call of `fun` this way
-      // ensures that any sync Error flows to the outer promise.
+  /*global Promise:false*/
+
+  /**
+   * The `promise` namespace contains utilities for working with the future-standard ES6 `Promise` API,
+   * using patterns not directly covered by it.
+   *
+   * Loading this module also loads the ES6 `Promise` shim module, `pentaho/shim/es6-promise`.
+   *
+   * @name pentaho.util.promise
+   * @namespace
+   * @amd pentaho/util/promise
+   * @ignore
+   */
+  return /** @lends pentaho.util.promise */{
+    /**
+     * Wraps a function call within a promise.
+     *
+     * The function is called **synchronously**.
+     * However, the returned promise is only settled asynchronously.
+     *
+     * The promise is fulfilled with the function's return value
+     * or rejected with an error thrown by the function.
+     *
+     * @param {function() : ?any} fun The function to call.
+     * @param {?Object} [ctx] The object on which to call `fun`.
+     *
+     * @return {Promise.<?any>} A promise for the function's return value.
+     */
+    wrapCall: function(fun, ctx) {
+      if(!fun) throw error.argRequired("fun");
+
       return new Promise(function(resolve) {
         resolve(fun.call(ctx));
       });
     },
 
+    /**
+     * Gets a promise for one or more modules, given their ids.
+     *
+     * When `deps` is a `string`, a single module is required,
+     * and the promise is fulfilled directly with the value of that module.
+     *
+     * When `deps` is an `Array`, zero or more modules are required,
+     * and the promise is fulfilled with an array with the values of those modules.
+     *
+     * Optionally, receives a contextual `require` function,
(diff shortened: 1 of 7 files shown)


test-js/unit/pentaho/util/promise.Spec.js
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        expectIt([]);
+        expectIt([null]);
+        expectIt([undefined]);
+      });
+
+      it("should fulfill to a single value when `deps` is a string", function(done) {
+        var mid = "pentaho/tests/promise/require/test1";
+        require.undef(mid);
```
carlosrusso: Isn't it a bit dangerous to add/remove amd modules in the test environment?
It seems that a sloppy test might have a non-local effect which might be hard to trace.
Could some of the tests be moved to a dependency injector such as [Squire.js](https://github.com/iammerrick/Squire.js) ?

test-js/unit/pentaho/util/promise.Spec.js
```diff
@@ -0,0 +1,316 @@
...
+          require.undef(mid1);
+          require.undef(mid2);
+          done();
+        }, done.fail);
+      });
+
+      it("should use the specified `localRequire` instead of the global one when `deps` is a string", function() {
+        // Executed synchronously
+        var localRequire = jasmine.createSpy("localRequire")
+            .and.callThrough(function(deps, calllback, errback) {
```
carlosrusso: The deliberate misspelled argument `calllback` can be easily missed, and might lead to unnecessary wasting of time.
What about renaming the argument to `f`?

package-res/resources/web/pentaho/util/promise.js
```diff
@@ -14,27 +14,101 @@
...
+     * and returns a new, identical promise for others to follow.
+     *
+     * @param {Promise} promise The promise to wait for.
+     * @param {function} fun The function to call.
+     * @param {?object} [ctx] The object on which to call `fun`.
+     *
+     * @return {Promise} An identical promise.
+     */
     "finally": function(promise, fun, ctx) {
+      if(!promise) throw error.argRequired("promise");
```
carlosrusso: I'm having trouble justifying this function, because it's not much more than a `.then` .
Can you point me to a place it is being used or describe your intentions behind this function?

carlosrusso:
(no more comments on this module for now)

test-js/unit/pentaho/util/promise.Spec.js
dcleao: That's an issue I haven't been able to solve well yet. I've tried using Squire a number of times — actually every time I had to test async stuff with AMD modules in pentaho/type. The problem is that it goes completely crazy and totally unrelated tests start breaking everywhere. I was able to use Squire in geo, I remember, but in common-ui, I've been unsuccessful. So I turned to doing it directly.
For simple cases, that don't have dependencies, this method has no hidden dangers.
The alternative would be to create requirejs contexts ourselves, but we would be leaking memory that way, something which come become a problem, and doesn't seem right any way.
The best solution would be one in which the requirejs library itself would provide a testing library.

test-js/unit/pentaho/util/promise.Spec.js
dcleao: Deliberate? No, just dyslexia or something like that. I'll fix it.

package-res/resources/web/pentaho/util/promise.js
dcleao: The only case in which it is being used now is in: https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/pentaho/visual/Wrapper.js#L948.

This is code to be removed in a near future — the VizAPI 2.5.
Honestly, I didn't think much on whether it was worth keeping it.
It's the promise counterpart to the finally of a try/catch/finally statement. You pass in a function that gets called in any case, success or failure. A promise is returned that does not change in any way the original promise, thus preserving the return value or failure reason to subsequent handlers.

This will typically be used for cleaning up state that should not become invalid in case of errors that are expected to possibly occur.

After writing all this, I remembered checking out the great `Q.js` version: https://github.com/kriskowal/q/wiki/API-Reference#promisefinallycallback.

Not that this affects our need to keep the code.

dcleao: Added comment saying we should consider removal of this utilityin the future.

test-js/unit/pentaho/util/promise.Spec.js
dcleao: Fixed :-)

dcleao:
@carlosrusso, @graimundo please re-review.

graimundo:
Looks good.

PR #625, opened 2016-02-10
[BACKLOG-6288] arg spec and doc

scottyaslan:
@pentaho/millenniumfalcon

### DIFF
--- a/package-res/resources/web/pentaho/util/arg.js
+++ b/package-res/resources/web/pentaho/util/arg.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,23 +20,71 @@ define([
 
   var A_slice = Array.prototype.slice;
 
+  /**
+   * @name pentaho.util.arg
+   * @class
+   * @amd pentaho/util/arg
+   *
+   * @classDesc Util methods for keyword args.
+   *
+   * @description Performs actions on keyword args.
+   */
   return {
+    /**
+     * Gets the value of an optional property of an object. The property is considered specified when its value is not a {@link Nully} value.
+     * When the property is not specified, the value of `defaultValue` is returned. The latter defaults to `undefined`.
+     *
+     * @param {?object} obj The object from which to get a property.
+     * @param {string} propName The name of the property.
+     * @param {any} [defaultValue] The default value.
+     *
+     * @return {any} The value of the property. If the property does not exist, returns `defaultValue`.
+     */
     optional: function(o, p, dv) {
       var v;
       return o && (v = o[p]) != null ? v : dv;
     },
 
+    /**
+     * Gets the value of an defined property of an object. The property is considered specified when its value is not `undefined`.
+     * When the property is not specified, the value of `defaultValue` is returned. The latter defaults to `undefined`.
+     *
+     * @param {?object} obj The object from which to get a property.
+     * @param {string} propName The name of the property.
+     * @param {any} [defaultValue] The default value.
+     *
+     * @return {any} The value of the property. If the property does not exist, returns `defaultValue`.
+     */
     defined: function(o, p, dv) {
       var v;
       return o && (v = o[p]) !== undefined ? v : dv;
     },
 
+    /**
+     * Gets the value of an required property of an object. The property is considered specified when its value is not a {@link Nully} value.
+     *
+     * @param {?object} obj The object from which to get a property.
+     * @param {string} propName The name of the property.
+     * @param {?string} pscope The name of the argument where the `obj` is received in the caller.
+     *
+     * @return {any} The found required property value.
+     * @throws {Error} Argument required. The `obj` must contain the `propName`.
+     */
     required: function(o, p, pscope) {
       var v;
       if(o && (v = o[p]) != null) return v;
       throw error.argRequired((pscope ? (pscope + ".") : "") + p);
     },
 
+    /**
+     * Slices the provided array.
+     *
+     * @param {Object} args Array of anything.
+     * @param {number} [start=0] The index of the `args` array to begin the slice.
+     * @param {number} [end] The index of the `args` array to end the slice at.
+     *
+     * @return {Object} Array containing the elements from the `args` array between the `start` and the `end`.
+     */
     slice: function(args, start, end) {
       switch(arguments.length) {
         case 0: throw error.argRequired("args");
--- a/test-js/unit/pentaho/util/arg.Spec.js
+++ b/test-js/unit/pentaho/util/arg.Spec.js
@@ -1,0 +1,163 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define([
+  "pentaho/util/arg",
+  "pentaho/util/error"
+], function(arg, error) {
+
+  "use strict";
+
+  /*global describe:true, it:true, expect:true, beforeEach:true*/
+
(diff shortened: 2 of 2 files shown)


package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,77 @@ define([
...
+  /**
+   * @name pentaho.util.arg
+   * @class
+   * @amd pentaho/util/arg
+   *
+   * @classDesc Util methods for keyword args.
+   *
+   * ### AMD
+   *
+   * Module Id: `pentaho/util/arg`
```
carlosrusso: The @amd tag has been merged, take a look

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,77 @@ define([
...
+   * @description Performs actions on keyword args.
+   */
   return {
+    /**
+     * Provides the optional property value of a given object or the given default value if the given object does
+     * not contain the sought after property value.
+     *
+     * @param {Object} o Keyword arguments.
+     * @param {String} p Property name to find in o param.
+     * @param {Object|String|Number} dv Default value to return if p param is not found.
```
carlosrusso: I think it would be more readable if we used more legible argument names, e.g. 
@param {Object} keyArgs -  Keyword arguments.
...

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,77 @@ define([
...
       var v;
       if(o && (v = o[p]) != null) return v;
       throw error.argRequired((pscope ? (pscope + ".") : "") + p);
     },
 
+    /**
+     * Slices the provided array
+     *
+     * @param {Object} args Keyword arguments.
+     *
```
carlosrusso: missing description for start, end arguments

scottyaslan: doh!

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
+   */
   return {
+    /**
+     * Provides the optional property value of a given object or the given default value if the given object does
+     * not contain the sought after property value.
+     *
+     * @param {?object} [obj] The object to search for given propName param.
+     * @param {String} [propName] Property name to find in obj param.
+     * @param {any} [defaultValue] The default value to return if propName param is not found. The default value of this param is undefined and is considered specified when its value is not a nully value. {@link Nully}
+     *
```
carlosrusso: I doubt that obj and propName are optional parameters...

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
+    /**
+     * Provides the required property value of a given object.
+     *
+     * @param {?object} [obj] The object to search for given propName param.
+     * @param {String} [propName] Property name to find in obj param.
+     * @param {?string} [pscope] The name of the argument where the obj param is received in the caller.
+     *
+     * @return {any} The found required property value.
+     * @throws {Error} Argument required. The o param must contain the p param property.
+     */
```
carlosrusso: the obj parameter must contain a property named propName

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
       throw error.argRequired((pscope ? (pscope + ".") : "") + p);
     },
 
+    /**
+     * Slices the provided array.
+     *
+     * @param {Object} [args] Array of anything.
+     * @param {Number} [start] The index of the args array to start the slice.
+     * @param {Number} [end] The index of the ars param array to end the slice at.
+     *
```
carlosrusso: args is not an optional parameter.
start is optional with a default value of 0
end is optional, defaults to...

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
 
+  /**
+   * @name pentaho.util.arg
+   * @class
+   * @amd pentaho/util/arg
+   *
+   * @classDesc Util methods for keyword args.
+   *
+   * The AMD module returns the arg utils.
+   *
```
carlosrusso: use the @amd tag

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
+    /**
+     * Provides the defined property value of a given object or the given default value if the given object does
+     * not contain the sought after property value.
+     *
+     * @param {?object} [obj] The object to search for given propName param.
+     * @param {String} [propName] Property name to find in obj param.
+     * @param {any} [defaultValue] The default value to return if propName param is not found. The default value of this param is undefined and is considered specified when its value is not a nully value. {@link Nully}
+     *
+     * @return {any} The found defined property value (a property is considered defined if its value is not undefined) or the default value.
+     */
```
carlosrusso: what about "Returns the value of the property. If the property does not exist, returns `defaultValue`." ?

package-res/resources/web/pentaho/util/arg.js
```diff
@@ -20,23 +20,73 @@ define([
...
+   *
+   * The AMD module returns the arg utils.
+   *
+   * @description Performs actions on keyword args.
+   */
   return {
+    /**
+     * Provides the optional property value of a given object or the given default value if the given object does
+     * not contain the sought after property value.
+     *
```
carlosrusso: This formulation is correct but it seems overly complex. Can you break it down to a simpler formulation?

package-res/resources/web/pentaho/util/arg.js
scottyaslan: I don't understand...isn't that what I did with: @amd pentaho/util/arg

package-res/resources/web/pentaho/util/arg.js
scottyaslan: ok, I see now, the [] imply the param is optional. What does the {?object} imply?

package-res/resources/web/pentaho/util/arg.js
carlosrusso: Sorry, I read the line 30 " The AMD module returns the arg utils." and assumed you forgot to add the @amd tag.
When you run "ant doc-js", doesn't that line add duplicate information?

package-res/resources/web/pentaho/util/arg.js
carlosrusso: take a look at http://usejsdoc.org/tags-param.html and http://usejsdoc.org/tags-type.html

dcleao: So, `[obj]` — optional — means you can omit the argument altogether, like calling it this way: `arg.optional()`. When one argument is optional, all arguments to its right should be optional as well.
And `{?object}` — nullable — means that the value of the argument may be null (we would actually want to mean _nully_ in this case, but whatever):
```js
var keyArgs = null; 
var value = arg.optional(keyArgs, "prop");
```
In this function, I'd say that the only optional argument is `defaultValue`.
Also, please do not use `{String}`, but use `{string}` instead. The same applies to the other "simple" primitive classes: `Number` and `Boolean`. Instances of String, Number and Boolean are objects, and quite different from their primitive, typeof == {string, number, boolean} counterparts.
For functions and objects, it's a bit more complicated.

dcleao: Please, when writing the name of an argument, enclose it in backtick quotes: The argument `obj`.

dcleao: Please do not extend the text of argument descriptions much, like what you did with the `defaultValue`.
It's generally preferable to have a paragraph in the main description, describing the behavior of the function, even if the paragraph focus on a single argument.
The description of arguments, should focus more on immediate and obvious stuff...
Something like this:
```js
/**
 * @param {?object} obj The object from which to get a property.
 * @param {string} propName The name of the property.
 * @param {any} [defaultValue] The default value.
```

package-res/resources/web/pentaho/util/arg.js
dcleao: Same notes as above.

package-res/resources/web/pentaho/util/arg.js
dcleao: end probably defaults to the length of the array and there's no way to denote that in JsDocs.

package-res/resources/web/pentaho/util/arg.js
dcleao: same notes as above.

package-res/resources/web/pentaho/util/arg.js
dcleao: Following the links that Carlos shared, you'll find the closure compiler type expressions link: https://developers.google.com/closure/compiler/docs/js-for-compiler#types, which is the "reference" for type expressions.

package-res/resources/web/pentaho/util/arg.js
davidmsantos90: Yes, the @amd tag already displays that information.

dcleao:
All fine for me :+1:

PR #626, opened 2016-02-11
[BACKLOG-6284] Base.js - Unit testing and dev API reference documentation

nantunes:
For review. @pentaho/millenniumfalcon

### DIFF
--- a/package-res/resources/web/pentaho/lang/Base.js
+++ b/package-res/resources/web/pentaho/lang/Base.js
@@ -382,7 +382,9 @@ define([
 
   //region Method Override
   function methodOverride(value, baseValue, rootProto) {
-    if(!value || !baseValue) return value || baseValue;
+    if(!value) return baseValue;
+
+    if(!baseValue) baseValue = inst_base;
 
     // Get the unwrapped value.
     var method = value.valueOf();
--- a/package-res/resources/web/pentaho/lang/Base.js
+++ b/package-res/resources/web/pentaho/lang/Base.js
@@ -67,26 +67,64 @@ define([
 
   // -----------
 
+  /**
+   * @private
+   *
+   * Defines the Base class.
+   *
+   * @returns {pentaho.lang.Base}
+   */
   function base_create() {
-    // The Base.Object root class is the base class for regular Object classes.
     var Base = base_root(Object, {}, "Base.Object");
     Base.version = "2.0";
+
+    /**
+     * @public
+     * @static
+     *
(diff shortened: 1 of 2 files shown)


nantunes:
For reference, initial review of @dcleao on a rewritten commit: https://github.com/nantunes/pentaho-platform-plugin-common-ui/commit/d326e1343ce75142cff66a7c8e2fc61b4defde48

nantunes:
Coverage info:
> Statements: 100% (173 / 173)
> Branches: 93.07% (94 / 101)
> Functions: 93.55% (29 / 31)
> Lines: 100% (162 / 162)

nantunes:
@dcleao Changes according to your comments.

Didn't create a jsdoc-only IBase interface, don't think it solves our problems. Any comment?

nantunes:
Unexplored cases:
- [ ] Override a property with a getter and/or setter (either don't permit or resolve how to provide an useful this.base.

dcleao:
@nantunes Hadn't seen the clone array bullet you added above...

In what an `IBase` interface is concerned, for moments, I thought you could be right, cause almost all of the Base interface is static.

But then, for my dismay, I found out that you can declare both **static** and **instance** members in an **interface** (by the way, @diogofscmariano, check out this case on static being legal/possible on arrays besides classes...):

idisposableinterfacewithstaticmembers

It's as easy as making something a _member_ of the interface directly, instead of of its "prototype":
```js
/**
 * @name disposeStatic
 * @memberOf pentaho.lang.IDisposable
 */
```

Interestingly, _TypeScript_ does have the concept of interfaces for the instance and for the static side of classes (checkout "Difference between static/instance side of class" a bit below [here](http://www.typescriptlang.org/Handbook#interfaces)).

One of the problems with the approach you took is that you're documenting `Base.Object` as being a subclass of `Base`, when, actually, `Base === Base.Object`. `Base.Object` is instead a subclass of the native `Object`.
Similarly, `Base.Array` is not a subclass of `Base`, but of the native `Array`.
(this description ignores the boot class which is created for each root, and actually creates a middle-man in the prototype chain; but this is a private implementation detail :-) )

nantunes:
@dcleao and @graimundo please review.

nantunes:
Review commit https://github.com/nantunes/pentaho-platform-plugin-common-ui/commit/43101a742dd73a9eedf7fd1fb9e874fae9aba7b8 for changes since last review.

graimundo:
Great Work!

PR #628, opened 2016-02-11
[BACKLOG-6289] DEV: Create unit tests and documentation for /pentaho/util/error.js

dcleao:
* `argInvalidType` now returns a standard JavaScript `TypeError`
* `argOutOfRange` now returns a standard JavaScript `RangeError`
* all errors now have an overridden `Error#name` property that shows up
  by standard behavior when errors converted toString; this now carries a string
  that denotes the error type, like: "ArgumentRequiredError", "OperationInvalidError", ...
* all error instances now have extra metadata properties like `argument` to contain the
  argument name of an `argXYZ` method, and `expectedTypes` that contains the array of
  expected type names of an `argInvalidType` error
* Added a kind of an error mock for use in specs — error.Match.js
  * Replaces direct use of `error` methods to test errors thrown by other code
  * `errorMatch` has one method per original `error` method, like
    `errorMatch.argRequired("foo")`
  * The methods, however, do not include arguments that are purely textual (like `reason`),
    as those are not matched against
  * The matcher works by creating a fake error instance that knows how to compare itself to
    real Errors, by using jasmine's `asymmetricMatch` [feature](http://jasmine.github.io/edge/introduction.html#section-Custom_asymmetric_equality_tester).
    Can be used as the value of matchers like `toEqual(.)` and `toThrow(.)`
    (but not `toThrowError`...)
  * We can take this technique further and apply it to the validation errors we have,
    to avoid having to include bundles to have proper error testing
  * On the downside, for each original error method, there's a corresponding error matcher class...
* Fixing paths in nodojo.karma.conf.js for the nth time...

@pentaho/millenniumfalcon what do you think of this as an alternative to direct use of the error factories (and, possibly, message bundles) for asserting errors?

### DIFF
--- a/config/nodojo.karma.conf.js
+++ b/config/nodojo.karma.conf.js
@@ -15,8 +15,8 @@ module.exports = function (config) {
       "config/context-begin.js",
 
       // SOURCE files
-      {pattern: "build-res/module-scripts/**/{*.js,*.html,*.xml}", included: false},
-      {pattern: "package-res/resources/web/**/{*.js,*.html,*.xml,*.properties}", included: false},
+      {pattern: "build-res/module-scripts/**/*.+(js|html|xml)", included: false},
+      {pattern: "package-res/resources/web/**/*.+(js|html|xml|properties)", included: false},
 
       // AMD configuration
       "build-res/requireCfg-raw.js",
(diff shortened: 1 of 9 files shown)


dcleao:
@graimundo updated with your feedback.

PR #640, opened 2016-02-16
[CLEANUP] Improved testability of pentaho/util/object.js

carlosrusso:
I could not run ant-js on my machine at the moment due to some problem with PhantomJS, please confirm that the tests pass.

### DIFF
--- a/package-res/resources/web/pentaho/util/object.js
+++ b/package-res/resources/web/pentaho/util/object.js
@@ -258,6 +258,11 @@ define(function() {
         });
 
       return Class.apply(inst, args || A_empty) || inst;
+    },
+
+    _forTestingPurposesOnly: {
+      setProtoCopy: setProtoCopy,
+      setProtoProp: setProtoProp
     }
   };
 
@@ -281,7 +286,7 @@ define(function() {
    */
   function copyOneDefined(to, from, p) {
     var pd = getPropertyDescriptor(from, p);
-    if(pd && pd.get || pd.set || pd.value !== undefined)
+    if(pd && (pd.get || pd.set || pd.value !== undefined))
       Object.defineProperty(to, p, pd);
     return to;
   }
@@ -298,6 +303,13 @@ define(function() {
   }
 
   function setProtoCopy(o, proto) {
+    if(proto === null){
+      // We can't set the prototype directly, so we need to "overwrite" all of its functionality
+      for(var q in o)
+        if(!o.hasOwnProperty(q))
+          o[q] = undefined;
+    }
+
     for(var p in proto) copyOneDefined(o, proto, p);
     return o;
   }
--- a/test-js/unit/pentaho/util/object.Spec.js
+++ b/test-js/unit/pentaho/util/object.Spec.js
@@ -451,56 +451,62 @@ define([
 
     }); // getPropertyDescriptor
 
-    describe("`setPrototypeOf` -", function() {
-      var Spam, protoEggs, parrot;
-      beforeEach(function() {
-        Spam = function() {
-          this.bar = "bar";
-        };
-        Spam.prototype = new function() {
-          this.spam = "spam";
-        };
-        parrot = new Spam();
+    [
+      {setPrototypeOf: O.setPrototypeOf, _label: "ES5"},
+      {setPrototypeOf: O._forTestingPurposesOnly.setProtoCopy, _label: "setProtoCopy"},
+      {setPrototypeOf: O._forTestingPurposesOnly.setProtoProp, _label: "setProtoProp"}
+    ].forEach(function(O) {
+      describe("`setPrototypeOf` - " + O._label + " variant -", function() {
+        var Spam, protoEggs, parrot;
+        beforeEach(function() {
+          Spam = function() {
+            this.bar = "bar";
+          };
+          Spam.prototype = new (function() {
+            this.spam = "spam";
+          })();
+          parrot = new Spam();
 
-        var myProto = function() {
-          this.spam = "eggs";
-        };
-        protoEggs = new myProto();
-      });
-
-      it("should return the input object, if the desired prototype is an object", function() {
-        expect(O.setPrototypeOf(parrot, protoEggs)).toBe(parrot);
-      });
-
-      it("should return the input object, if the desired prototype is `null` ", function() {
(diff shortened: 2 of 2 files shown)


carlosrusso:
I finally managed to get PhantomJS to run. Ran  `ant test-js` successfully.

nantunes:
Don't really like exposing the private methods... See https://github.com/carlosrusso/pentaho-platform-plugin-common-ui/pull/1 for an alternative proposal.

But now the `make - should pass the array of arguments to the constructor` test is failing... Can you help me?

dcleao:
Does "use strict" even let you change `Object.prototype.__proto__`. Anyway, even if it lets you, by messing with that, you're messing with the source of standard JavaScript methods for many instantiated objects. The consequences are undefined...

nantunes:
The failing test is resolved. See updated https://github.com/carlosrusso/pentaho-platform-plugin-common-ui/pull/1.

@dcleao I'm now restoring the _normality_ sooner, before any test. Do it make more comfortable?

I understand your concerns, it's just an alternative to exposing the private methods.
@pentaho/millenniumfalcon any comments?

carlosrusso:
The interesting question this raises is really on how acceptable is it to expose private functions for the sake of code coverage.
There are four approaches apparently:
1. we do not care at all about code coverage on private code and we must take that into account when looking at code coverage
2. no private code is exposed, but the tests on the "public" interface attempt to trigger the execution of the underlying private code
3. the private code is somehow exposed, an the tests on the "public" interface also tests the private code
4. the private code resides be in a private module required by the public code and is covered by "private" tests on dedicated spec files.

dcleao:
A series or disparate opinions and my conclusion.

Exposing private stuff from modules for the purposes of testing is a typical, well accepted approach.

In this special case, the only way to exercise (and test) certain private functions is by using a different JS engine — we're testing something like a shim.

While externalizing the engine specific code would make it testable, it would also impact the code in ways that I don't think are worth it — it affects the normal case negatively: more modules and slower (loading more modules).

In this case, the private functions could exist as separate modules without change or loss of functionality. However, in many cases, this approach wouldn't be feasible, or would imply changing the code.
For example, instead of using a closure, a "stateless" function with more arguments would have to be used, or alternatively, a class, to hold such state.
Such changes, can negatively impact the code.

When the public API of a module provides all of the needed controls to exercise all of the code paths it contains, exposing a private method for testability is a matter of being able to target tests to important parts of its internals. It's a choice made for practical reasons.

Otherwise, if certain code paths depend on the environment where the code is run — as is the case of global variables and engine specific features — either: 
1. such environmental factors are mockable
2. are temporarily changeable (e.g. global variables)
3. the tests are run in every possible environment
4. internal parts that only run in certain environments are exposed somehow, to allow some testing in a single environment.

If the private functions that are exposed cannot affect the working of the module these cannot be used to exploit it. It's simply a matter of revealing unwanted details. Details that can change over time, without notice. It's the same as with private and protected methods of classes: you can use it, but we can't take responsibility for that.

In practical terms, it's generally desirable (and many times cheaper) to only make the codebase more complex when there are actual "business" needs, and use certain hacks to address certain quirks...

Now, imagine if a library had to externalize modules for every piece of engine fallback quirk there were, it would affect readability, size, performance...

In the end, the tests that are _really required_ — testing that when setPrototypeOf uses the fallback/copy behavior, the rest of the code still works — will still not be being tested...
And for these, we either make things much more complex in terms of the unit tests and the testing environment, or we accept that we should be running the tests in IE...

nantunes:
Another alternative proposal: https://github.com/carlosrusso/pentaho-platform-plugin-common-ui/pull/2

@graimundo The new module host capabilities and not the shim itself. `pentaho/util/object` is already intended to be used as a sort of shim, with `O.setPrototypeOf` serving as an abstraction layer.
So, it is a module to be used by shims. New checks should be added as needed.

@pentaho/millenniumfalcon please comment.

carlosrusso:
Seems to be a nicer approach






On Wed, Feb 17, 2016 at 8:15 AM -0800, "Nelson Antunes" <notifications@github.com> wrote:












Another alternative proposal: carlosrusso#2



@graimundo The new module host capabilities and not the shim itself. pentaho/util/object is already intended to be used as a sort of shim, with O.setPrototypeOf serving as an abstraction layer.

So, it is a module to be used by shims. New checks should be added as needed.



@pentaho/millenniumfalcon please comment.

dcleao:
@nantunes I like the environment capabilities module. It concentrates all environmental specific tests in a single, replaceable place. This solution seems to solve all of the stated problems (complexity, speed, testability). I don't like the code in the test file, though. You're getting too much into the privates of RequireJs... It's funny, though, cause we're trying to not expose our own private code, LOL.

dojo uses a similar thing, a generic module, "dojo/has", a dictionary of capabilities. I think it also works as a smart amd loader plugin, that allows conditional loading of other modules. AMD configuration might also allow to specify arbitrary flags.

I'm not sure the `pentaho/shim` folder is the right place. It is convenient, because it already exists, but it isn't a perfect match in what the module's function is concerned.
What do you think of `pentaho/util/has`? It could remain a private module, in the spirit of the `util` folder, at least initially.

carlosrusso:
+1 on renaming to pentaho/util/has.

carlosrusso:
On dependency injection : I would like to experiment with jasq, seems to fit our needs

pamval:
Please check with Architecture before experimenting with yet another library/framework.

nantunes:
@dcleao and @carlosrusso Ok, will rename the module and also try to find a less hacky way of cloning the require env. Will not do it today, tho.

@carlosrusso maybe you should close this PR and I will do another one (including your commit) latter?

carlosrusso:
@nantunes no, let's go ahead with this one.When we get an answer from Architecture on how to do dependency injection on tests, we'll amend this file.

carlosrusso:
on second thought, an approach that already includes the proposed "pentaho/util/has" module would be better.

nantunes:
I've created a separate PR for `pentaho/util/has` - https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/647.

As for the tests, I think we should follow @dcleao proposal: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/646.

PR #643, opened 2016-02-17
[BISERVER-12918] the report parameter/prompt panel flashes when autosubmit off

MikhailHubanau:
Exclude "" value from nullValueParams if it is TextInputComponent.

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -821,7 +821,7 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          * @param {String} name
          * @param {Object} value
          */
-        parameterChanged: function (param, name, value) {
+        parameterChanged: function (param, name, value, type) {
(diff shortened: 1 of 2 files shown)


package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -835,12 +835,15 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
             }
           }
 
-          if (!value || value == "" || value == "null") {
-            if (!this.nullValueParams) {
-              this.nullValueParams = [];
-            }
+          if(!(value == "" && type == "TextInputComponent"))
```
annapopova1: To avoid using additionally "type" input parameter we can check "param.list" property. Something like this:
```javascript
          if (param.list && (!value || value == "" || value == "null") {
            if (!this.nullValueParams) {
              this.nullValueParams = [];
            }
            this.nullValueParams.push(param);
          }
```
What do you think?

MikhailHubanau: I checked it works.

PR #644, opened 2016-02-18
[CLEANUP] Document thrown errors using the JsDocs @throws tag

dcleao:
@pentaho/millenniumfalcon please review.

This PR is based on the commit of #641, so please only merge after the latter.

PR #645, opened 2016-02-18
[CLEANUP] Create pentaho.lang.XyzError classes

dcleao:
* Made the error classes public, as they were showing up in public documented
  methods in any way.
* Left the private pentaho/util/error façade there, although now it is simply
  a mostly empty convenience for not having to require many Error class modules...
* Adapted the testing errorMatch utility
* Moved pentaho/util/error's inner methods to pentaho/util/text
* Added a new Base root for errors: Base.Error, that derives from the native
  Error class.

This PR depends on the previous PR https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/644, so don't merge before that one.

### DIFF
--- a/package-res/resources/web/pentaho/lang/Base.js
+++ b/package-res/resources/web/pentaho/lang/Base.js
@@ -76,7 +76,27 @@ define([
    * @returns {Class.<pentaho.lang.Base>}
    */
   function base_create() {
-    var Base = base_root(Object, {}, "Base.Object");
+    /**
+     * @classdesc `Base` Class for JavaScript Inheritance.
+     * Based on Base.js by Dean Edwards, and later edited by Kenneth Powers.
+     *
+     * @class
+     * @name Base
+     * @memberOf pentaho.lang
+     * @amd pentaho/lang/Base
+     *
+     * @description Creates a new object of `Base` class.
+     *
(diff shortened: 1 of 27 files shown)


package-res/resources/web/pentaho/lang/Base.js
```diff
@@ -153,6 +183,15 @@ define([
 
     // Used by BaseBoot.extend, just below
     BaseBoot.prototype.extend = inst_extend;
+
+    // ---
+
+    if(!baseConstructor) {
```
nantunes: I would externalise this constructor as `baseObjectConstructor`, like `baseErrorConstructor`, instead of having a default value. There is no problem changing the signature of the function, as it is private.

In the future we'll probably need a `baseArrayConstructor`, so it seems coherent.

nantunes:
All seems great to me, great improvement!

package-res/resources/web/pentaho/lang/Base.js
dcleao: Good idea!

dcleao:
Externalized the constructors.

graimundo:
Agreed, great improvements :)
Not sure about leaving pentaho/util/error as a "loading helper" for all the Error modules but I don't have a strong opinion against it either.

PR #646, opened 2016-02-19
Testing requirejs - DO NOT MERGE

dcleao:
This PR is for discussion only.
 
Hi, this is my attempt at understanding the minimal features that RequireJS would need to expose for being able to test with it properly. Check out require-test.js file and robject.Spec.js for an example using it.

Didn't test if it works in those scenarios where Squire and Jasq fails...

The fact that one of the tests still fails in PhantomJS — the Object.setPrototypeOf one —, tells us that for the env strategy to work, we can only trick the system if we're downgrading its features. We can't test the Object.setPrototypeOf branch in IE6, for example :-)...

How can we solve that problem? Skip the unit tests that an environment can't test? 

@pentaho/milleniumfalcon

### DIFF
--- a/test-js/unit/pentaho/util/robject.Spec.js
+++ b/test-js/unit/pentaho/util/robject.Spec.js
@@ -1,0 +1,115 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define(function() {
+  "use strict";
+
+  /*global describe:false, it:false, expect:false, beforeEach:false, beforeAll:false, Object:false*/
+
+  describe("squire pentaho.util.object - ", function() {
+    var Spam, protoEggs, parrot;
+
+    beforeEach(function() {
+      Spam = function() {
+        this.bar = "bar";
+      };
+      Spam.prototype = new (function() {
+        this.spam = "spam";
+      })();
+      parrot = new Spam();
+
+      var myProto = function() {
+        this.spam = "eggs";
+      };
+      protoEggs = new myProto();
+    });
+
+    it("should return the input object, if the desired prototype is an object", function(done){
+      var localRequire = require.new();
+
+      localRequire(["pentaho/util/object"], function(O) {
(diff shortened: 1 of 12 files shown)


dcleao:
@pentaho/millenniumfalcon

test-js/unit/pentaho/util/robject.Spec.js
```diff
@@ -0,0 +1,126 @@
...
+      has: {
+        "Object.setPrototypeOf": false,
+        "Object.prototype.__proto__": false
+      }
+    }].forEach(function(conf) {
+
+      it("should return the input object, if the desired prototype is an object - " + conf._label, function(done) {
+
+        // Mocks
+        define("pentaho/shim/env", conf);
```
nantunes: Aren't you changing permanently the value of `pentaho/shim/env`?

Wouldn't it be better to `define("mock/pentaho/shim/env", conf);` and create a map in the new context? I can be encapsulated in `newContext`, so the parameter wouldn't be just _syntax sugar_.

nantunes:
Overall, good solution.

test-js/unit/pentaho/util/robject.Spec.js
dcleao: Whatever you `define` just before creating a new require context gets captured into that context, so this doesn't affect the global registration. Alternatively, if this seems strange, the user can specify the "mocks" keyArgs. But there's really nothing we can do against this having this effect.
This also allows all of the features of `define`, comparing to the more limited mocks option.

On rethinking this, since last night, I think the `keyArgs` argument should instead be a full requireJS configuration, with an additional, customly handled property, "mocks".

dcleao: To support RequireJS behavior wrt to intaking, see: https://github.com/jrburke/requirejs/blob/master/require.js#L1438.

dcleao: Squire and Jasq use this same technique to implement mocks.

nantunes: Ok, wasn't aware of this behaviour. Makes sense, then.

dcleao:
Rebased on https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/647.

dcleao:
Updated to even more simple and requirejs aligned version.

test-js/unit/require-test.js
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Creates a new RequireJS context and returns its `require` function.
+   *
+   * The new context's initial configuration is that of the parent context.
+   *
+   * To apply additional configurations to the context, call the `config` method.
+   *
+   * To provide module definitions to the context, call the `define` method.
+   * Only non-anonymous methods are supported.
```
nantunes: `non-anonymous modules`, I suppose.

nantunes:
Great, I love it!

nantunes:
Maybe we can now rebase this and merge (excluding `robject.Spec.js`).

dcleao:
https://github.com/requirejs/requirejs/issues/1495

dcleao:
This has been merged as part of https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/711.

PR #647, opened 2016-02-19
[CLEANUP] Modularize engine dependent checks

nantunes:
Creates `pentaho/util/has`. To be expanded as needed. For now, used by `pentaho/util/object`.

### DIFF
--- a/package-res/resources/web/pentaho/util/has.js
+++ b/package-res/resources/web/pentaho/util/has.js
@@ -1,0 +1,31 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


carlosrusso:
Excellent!

dcleao:
Idem!

dcleao:
Two notes:
1. Change copyright year on new file
2. Would prefer `has` was a function: `has(capability: string) : boolean` and not exposing the internal hash.

graimundo:
+1 on `has` as a function

nantunes:
@graimundo and @dcleao Updated with `has` as a function.

dcleao:
It would be super cool if has("toString") wouldn't return true unless it is actually a capability :-)

nantunes:
Well... in a sense it is true that it supports `toString`. :smile: 

Updated.

nantunes:
Ups, forget it... Using one capability in the capabilities module... Will change.

nantunes:
There, history rewritten.

For the record, my previous stupid fix was:
```javascript
Object.setPrototypeOf(capabilities, null);
```
In a file whose sole purpose (for now) is to check if... `Object.setPrototypeOf` exists.

dcleao:
Agree that in a sense it supported toString :-)

dcleao:
It would be super cool if I could register a capability named hasOwnProperty!

nantunes:
Everything that makes you happy. 🖖🏻

dcleao:
Super cool :+1: !

PR #648, opened 2016-02-23
[BISERVER-13034] Implement UI: JS logic, markup

Aliaksandr-Kastenka:
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/dojo/pentaho/common/SplashScreen.html
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/dojo/pentaho/common/SplashScreen.js
was taken as basis.
@pamval @tmorgner 
Please review.


Aliaksandr-Kastenka:
@tmorgner 
So we will not use Feedback screen for pentaho-interactive-reporting?

tmorgner:
PIR can reference that screen directly from the report-viewer, in the same way it references the code of the report-viewer already.

PR #649, opened 2016-02-24
[BACKLOG-6757] - PIR - Human readable filters and Filters created by …

plagoa:
…prompts are now displaying (None)


plagoa:
The function "buildFilterText" is only called in PIR, no other plugins should be affected by this.

pamval:
I'm accepting this since Paulo is 100% sure this is only called from PIR.

I'll keep a bookmark of this pull request in case I need it for Paulo's performance review later in the year :-P

pamval:
Cherry-pick to 6.1 after validation, please

PR #655, opened 2016-03-01
[FIX][BACKLOG-6625] Normalize resourceIDs in i18nMock plugin

nantunes:
@dcleao please review.


dcleao:
I know we don't have strict formatting rules, but it's a good rule to preserve the active rules in a file, not to leave it in no man's land :-(
Otherwise, great!

nantunes:
Ok. _"Don't mess with the original"_ style rule seems ok to me. 😉

dcleao:
I know I sometimes failed to comply with this rule in the past... But, anyway, I try :-)

dcleao:
Great!

PR #661, opened 2016-03-03
[BACKLOG-6419] Prevented failures on the resolution of dependencies i…

carlosrusso:
…n the filter classes. Minor adjustments to documentation.


carlosrusso:
@pentaho/millenniumfalcon Please review. To solve the issue with dependency resolution, all accesses to the filter classes must come via the `pentaho.data.filter` module.

PR #663, opened 2016-03-03
[CLEANUP] Use the "mocha" karma reporter instead of the basic "progress" reporter.

dcleao:
This uses a mocha-like reporter for the DEV-time karma configs. It prints a nice outline of the test suites, marking them green, red or gray according to success, failure and ignored (https://www.npmjs.com/package/karma-mocha-reporter).

The only problem I see with the reporter is that console.logs get interleaved with the test titles. And code bases like prompting and angular directives abuse of console.log'ing... But this is a problem we already have...

What do you guys think?

@pentaho/millenniumfalcon, @pentaho/orlandocalisbon please review.

### DIFF
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -7,7 +7,8 @@ module.exports = function (config) {
     plugins: [
       'karma-jasmine',
       'karma-requirejs',
-      'karma-chrome-launcher'
+      'karma-chrome-launcher',
+      'karma-mocha-reporter'
     ],
 
     files: [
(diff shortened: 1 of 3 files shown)


carlosrusso:
Yes, please.

dcleao:
@pentaho/orlandocalisbon do you feel very strong for the "progress" reporter?

davidmsantos90:
Looks good and something like this would be great ... @carlosrusso already had converted me to the [karma spec reporter](https://www.npmjs.com/package/karma-spec-reporter), that is similar to this one and it is very helpful.

diogofscmariano:
Dont have a strong opinion. how does it integrates with Sonar?

dcleao:
The `karma.conf.ci.js` file is unchanged, and that is the configuration that has coverage activated.

dcleao:
So, I'd say we're just waiting for a button press.

PR #666, opened 2016-03-04
[BACKLOG-6898] - Re-implement PromptPanel#getString function

plagoa:
- Removed the getString function from the PromptPanel
- Required the localization service on the SubmitComponentBuilder
- Removed the unit tests from the prompt panel


plagoa:

annapopova1:
@plagoa what about properties files with localization strings?

PR #669, opened 2016-03-04
[BACKLOG-6535] Unit tests for instances' serialization

nantunes:
@dcleao Care to have a look and comment?

I have some deeper questions that is better to resolve in person, but this is a start.


nantunes:
Sorry, PR to the wrong repo.

PR #670, opened 2016-03-04
[CLEANUP] Renaming properties to conform to the is/has naming scheme

dcleao:


carlosrusso:
isAwesome

davidmsantos90:
@carlosrusso your property is undefined. It should be isAwesome = true
:+1:

PR #676, opened 2016-03-08
[BISERVER-13106]

diogofscmariano:
- Added auto complete component back
- Fixed BISERVER-12961 without having checks on the values array, creating unneeded updates
- Fixed param property, storing the param definition for the component. With the prompt flow, this property was not updated properly causing values to return to its original value
- Updated unit tests accordingly


diogofscmariano:

tmorgner:
The changed StaticAutocomplete/Text-component code is back to how it was in 5.4. So it should work as before.
(Not counting the much improved cleanup and test-coverage.)

krivera-pentaho:
Looks good to me

PR #680, opened 2016-03-09
[BACKLOG-6961] - Taking into account an ENVIRONMENT_CONFIG if defined…

jvelasques:
…, to easily setup test environments.


jvelasques:

PR #684, opened 2016-03-14
[BISERVER-1225] As an end user when executing a report run, I want fe…

Aliaksandr-Kastenka:
…edback on what is happening with the request(feedback screen updated according to the ux notes)

@tmorgner 
Please review.

### DIFF
--- a/package-res/resources/web/dojo/dijit/themes/pentaho/ProgressBar.css
+++ b/package-res/resources/web/dojo/dijit/themes/pentaho/ProgressBar.css
@@ -10,7 +10,7 @@
 
 .tundra .dijitProgressBarTile {
 	/* inner container for finished portion when in 'tile' (image) mode */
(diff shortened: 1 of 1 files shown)


tmorgner:
Great, you just changed the appearance of ALL progress bars in the pentaho platform. Good for you!

Now, back that change out, move the style definition to the report-viewer, and write a style-selector that works there (and only there). If your style-selector mentions the "feedbackScreen" ID, your local style rule will override the global one.

And let me repeat that for you, in case it was unclear the last time you tried to make changes to the global stylesheets:  
If you make changes in a shared project, every user of the shared project will use that changed code. If your code is aimed at a single component in a single module, DO NOT MAKE THAT CHANGE HERE! Keep that change local. 

If you do make the change globally, research what components are affected by it and then test how they react to your change. I'm quite sure that removing the background image from those other progress bars will have some visual impact there.

PR #690, opened 2016-03-17
[BACKLOG-5985] Added test and documentation to pentaho.data.filter.Ab…

carlosrusso:
…stractFilter#walk

It was convenient to add a  `walk` method to the filters in order to transform a given filter tree into another.

### DIFF
--- a/package-res/resources/web/pentaho/data/_filter/AbstractFilter.js
+++ b/package-res/resources/web/pentaho/data/_filter/AbstractFilter.js
@@ -87,9 +87,9 @@ define([
      * @param {?pentaho.data.filter~transformIteratee} iteratee - Function which will transform this filter.
      * @return {!pentaho.data.filter.AbstractFilter} Transformed filter.
      *
-     * @see pentaho.data.filter.AbstractTreeFilter#walk
+     * @see pentaho.data.filter.AbstractTreeFilter#visit
      */
-    walk: function(iteratee) {
+    visit: function(iteratee) {
       return iteratee(this, null);
     },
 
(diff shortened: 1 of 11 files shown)


package-res/resources/web/pentaho/data/_filter/AbstractFilter.js
```diff
@@ -79,9 +79,20 @@ define([
       return _apply(this, dataTable);
     },
 
-    transform: function(iteratee){
+    /**
```
dcleao: Not sure `walk` is the usual term for a traversal function that rewrites or transforms the tree. I saw this on MSDN: https://msdn.microsoft.com/en-us/library/bb882521(v=vs.90).aspx. Please take a look, even seeing if the iteratee contract is similar to ours, and if any better method exists :-)

lucboudreau:
http://10.100.11.222:8080/job/wingman-osgi/43/

lucboudreau:
Error building command: /home/devuser/docker-build/assembly/build/assembly-0.0.1/source/pentaho-platform-plugin-common-ui-base-690-15e04804-7d1d-44f1-a9b9-5ec8a68e5d3b/pom.xml (No such file or directory)

package-res/resources/web/pentaho/data/_filter/AbstractFilter.js
carlosrusso: Since the idea here is to return a transformed version of a node (tranversing the tree if need be), I'll change the method's name to "visit"

dcleao: Not that "visit" gives that idea, solely by the meaning of the word — as your statement seems to enunciate — it just looks like it is a common term for such an operation.

carlosrusso:
@pentaho/millenniumfalcon Renamed #walk to #visit and rebased. Minor touches to documentation.

PR #693, opened 2016-03-18
[BACKLOG-5799] Report Viewer - will not show a special pop-up if ther…

annapopova1:
…e is a fatal error

Fixed 2 problems:
- incorrect code for setting style to elements. "dojo/dom-style" works with single domNode but "dojo/query" returns nodeList and we should use forEach to set width.
- progress indicator from blockUi is located over the dijit dialog (pentaho.common.Dialog). he reason is blockUi uses 1000 and 1011 z-index by default but dijit dialog uses 950 z-index by default. To fix it I set default z-index for dijit dialog.

### DIFF
--- a/package-res/resources/web/dojo/pentaho/common/Dialog.js
+++ b/package-res/resources/web/dojo/pentaho/common/Dialog.js
@@ -32,6 +32,10 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
             closeIcon: undefined,
             _onCancelCallback: undefined,
 
+            constructor: function() {
+              Dialog._DialogLevelManager._beginZIndex = 1051;
+            },
+
             setLocalizationLookupFunction: function(f) {
               this.getLocaleString = f;
               this._localize();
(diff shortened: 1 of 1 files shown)


package-res/resources/web/dojo/pentaho/common/Dialog.js
```diff
@@ -32,6 +32,10 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
             closeIcon: undefined,
             _onCancelCallback: undefined,
 
+            constructor: function() {
+              Dialog._DialogLevelManager._beginZIndex = 1051;
```
diogofscmariano: Cant we set this via css just for RV/PIR?

annapopova1: Looks like we can't.
For dijit/Dialog z-index property is hard-coded in js code (950 level). It overrides all css z-index properties.

krivera-pentaho:
This looks good to me, since you answered diogo's question.

diogofscmariano:
I'm really afraid with this question. Anna, can you please search where this is being used on the platform in order to know what are the tests we need to do?

annapopova1:
Investigation of usages:
common-ui
	pentaho/common/Dialog is parent for:
		pentaho.common.datasourceselect
		pentaho.common.FilterDialog
		pentaho.common.MessageBox
		pentaho.common.SplashScreen
		pentaho.common.TemplatePicker
		
RV
	pentaho.reportviewer.ReportDialog - do not find where is used
	pentaho.common.MessageBox
	
PIR
	pentaho.pir.PageSetupDialog
	pentaho.common.datasourceselect
	pentaho.common.FilterDialog
	pentaho.common.MessageBox
	pentaho.common.SplashScreen
	pentaho.common.TemplatePicker

PR #701, opened 2016-03-21
[TEST] - When dealing with JSON, we can't rely on a fixed order, refa…

jvelasques:
…ctoring tests to not use JSONObject.ToString as a comparison


jvelasques:
@pamval @diogofscmariano these two tests were failing in my machine, I noticed it was due to the json String order, which was different than the expected. The refactored tests should pass no matter the order.
please review

PR #702, opened 2016-03-22
[BACKLOG-7199] Fatal error occurs in case async report generation and…

dzmitry-hryhoryeu:
Fatal error occurs in case async report generation and empty page number text box

Thomas, take a look please.
@tmorgner


tmorgner:
This is a more general error, and is actually not related to the async processing. The same error occurs in a normal 6.0 installation. Seems that Dojo's isValid does not check for empty strings.

PR #703, opened 2016-03-22
[BACKLOG-6961] Fixed usage of underscore dependency

afrjorge:
- Removed duplicate path configuration for the same underscore file
    - Added a mapping to guarantee backwards compatibility


afrjorge:

PR #704, opened 2016-03-22
[CLEANUP] Including logger util and updating license headers

davidmsantos90:

### DIFF
--- a/package-res/resources/web/pentaho/util/logger.js
+++ b/package-res/resources/web/pentaho/util/logger.js
@@ -1,0 +1,112 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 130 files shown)


package-res/resources/web/pentaho/util/logger.js
```diff
@@ -0,0 +1,127 @@
...
+    error: function (m) {
+      return _log(m, "error");
+    },
+
+    /**
+     * Logs a message at exception level.
+     *
+     * @param {string} m - Message to log.
+     */
+    exception: function (m) {
```
dcleao: Forgot this, but it would be good to remove the "exception" method to prevent its uses.

davidmsantos90:
@dcleao @graimundo removed exception function

package-res/resources/web/pentaho/util/logger.js
```diff
@@ -0,0 +1,112 @@
...
+   */
+  var logger =/** @lends pentaho.util.logger */{
+
+    /**
+     *  Property enumerating the various log levels.
+     *
+     *  @property logLevels
+     *  @type Array
+     */
+    logLevels: ["debug", "log", "info", "warn", "error", "exception"],
```
graimundo: You removed the function but forgot to remove the level. Remove the exception loglevel.

PR #705, opened 2016-03-22
[BISERVER-12918] - _areParamsDifferent should also compare String arrays

jvelasques:


jvelasques:

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -354,6 +354,10 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
               return paramValue.toUpperCase() != paramSelectedValue.toUpperCase();
             case "java.sql.Date": // Set time to zero to eliminate its influence on the days comparison
               return (new Date(paramValue).setHours(0,0,0,0)) != (new Date(paramSelectedValue).setHours(0,0,0,0));
+            case "[Ljava.lang.String;": // Comparing String arrays
```
diogofscmariano: is this the toString of the type we receive?

diogofscmariano:
@jvelasques can we have tests covering this?

package-res/resources/web/prompting/PromptPanel.js
tmorgner: Yes

diogofscmariano: So it is safe to use this for arrays?
Is it worth it to have something for other arrays or this is the only use case we have?

tmorgner: This only handles "Steelwheels" cases (ie you have no real data model and all column keys are strings). I'd rather see a generic array test instead. If paraType starts with "[" we know we have an array. The members of the array will be strings (thanks to our XML parser), so the comparison is still valid as it is. But now we also handle integer-arrays and all other common types used in database keys.

tmorgner: The reporting engine and server side parameter handling guarantees arrays as types for multi-select parameters.  
All multi-select parameters expect their values as arrays and will not accept creepy modern stuff like Lists, Sets or so on. 

Likewise with contents of the arrays - we get them from the server as strings, and as we only perform an equals comparison, we can treat them as black-box values. As long as they compare correctly we will handle them correctly.

jvelasques: thanks @tmorgner for the info, will change this to handle all array types then :)

jvelasques:
@pentaho/orlandocalisbon added unit test, handling any array type, please review

PR #707, opened 2016-03-23
[BISERVER-13107] Query-driven textbox parameters defined as Number ar…

SergeyTravin:
…e formatted as ##,### when screen reloads

- remove default formating by dojo

### DIFF
--- a/package-res/resources/web/prompting/components/StaticAutocompleteBoxComponent.js
+++ b/package-res/resources/web/prompting/components/StaticAutocompleteBoxComponent.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


SergeyTravin:
@diogofscmariano, @krivera-pentaho, @rfellows According the e-mail from Thomas Morgner  we should not use format exclude than its define. If we do not have  formatter than we use v.label.   Could you please review it?

SergeyTravin:
@diogofscmariano, @krivera-pentaho, @rfellows Could you please review it?

krivera-pentaho:
This looks good to me.

diogofscmariano:
are we aware that removing this code section will probably break http://jira.pentaho.com/browse/PIR-920?

tmorgner:
No, I disagree. Our bug here was that we applied a formatter by "guessing" the format string. We applied a format to anything that looks remotely like a number, regardless of whether the parameter had a format-string defined. We now make that more strict and only format data if the user explicitly requested it.

It therefore should NOT break PIR-920, assuming that their prompt has an explicit format-string defined.

 If they have defined the parameter with a format-string, we will still apply the formatter, according to this line:

> initialValue = this.formatter ? this.formatter.format(this.transportFormatter.parse(v.label)) : v.label;

diogofscmariano:
Thanks for the clarification Thomas

PR #709, opened 2016-03-23
[CLEANUP] Multiple cleanup issues done in the context of BACKLOG 6052 (II)

dcleao:

### DIFF
--- a/package-res/resources/web/pentaho/lang/Base.js
+++ b/package-res/resources/web/pentaho/lang/Base.js
@@ -35,10 +35,17 @@
  * 11. Class.valueOf removed. No longer needed... ?
  * 12. `Base#__root_proto__` is a new constant, non-enumerable property, set at each "Base" root prototype.
  * 13. `Base#__init__` is a new constant, non-enumerable, property, set at the prototype of "Base" classes
- *      that have an own initialization method (specified with the `constructor` property).
+ *     that have an own initialization method (specified with the `constructor` property).
  * 14. Added new Class.implementStatic method to allow to specify the class interface after extend.
  *     Is parallel to Class.implement.
  * 15. Any existing static methods are inherited (not only standard Base ones).
+ * 16. Dropped support for the overload Base#extend(name, value).
+ *     The same method now supports only the signature Base#extend(instSpec[, keyArgs]).
+ * 17. Added support to augment the set of per-class, inherited property names that are excluded from
+ *     instance-extension operations.
+ *     Specify `extend_exclude` when sub-classing, e.g. `Base.extend({extend_exclude: {a: true, b: true}});`.
+ * 18. Added support to control, per-class, the properties extension order in instance-extension operations.
+ *     Specify `extend_order` when sub-classing, e.g. `Base.extend({extend_order: ["b", "a"]});`.
  */
 define([
   "../util/object",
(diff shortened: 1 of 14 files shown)


lucboudreau:
Please don't use CLEANUP if you have a lot of commits to push, especially since there is a case number to relate these to in the first place. These numerous cleanup commits fill history with untraceable changes. The case # is indeed within the PR name, but this does not make it to the history.

dcleao:
If they were directly related to the issue BACKLOG 6052, they would be marked with it.
These are refactorings done _a propo_ of that issue, but are not the direct goal of the issue.
We usually do this type of cleanups between sprints, for small-changes that aren't registered as issues.

We can review this practice. As I understand this, this being a long-term development effort, capturing every single cleanup commit as a story is not so important — you cannot assume individual commits are cherry-pickable as is usual in maintenance projects.
We could register them in a nearby issue being worked on... but that's not ideal as well.

In what having a single PR for multiple CLEANUP commits is concerned.
I don't see the problem with having  multiple CLEANUP commits if they handle different or unrelated things. In our team, we've been trying to not squash everything till the last drop, as it results in incomprehensible commit lists. We do squash if commits are a sequence of refinements, and the path to get to the final result is irrelevant. But here, this is already the result of squashing each section/topic to its final state.

Should I issue one PR per cleanup commit?

lucboudreau:
I see. I had not realized that each of these commits cover multiple other cleanups. In the case here, let's leave them as is. They are big enough individually as is. Do not squash them further.

Concerning commit messages however, I still think you should add the case number to its prefix. If need be, create a Jira case just for the cleanup commits, and assign it to the right release. Joe and his team have to reconcile Jira roadmaps with Git commits, and abusing CLEANUP does not help them. Plus, if we need to move these changes to another branch, we won't have to comb through all of the CLEANUP messages to figure out which of the CLEANUP commits are necessary; we can just search for a case number in commit messages and get them all out at once.

dcleao:
Ok, I will tag them with the nearest issue, then — this will tie the issues to a release.

About _moving_ changes to other branches, I don't think you can generally and safely cherry-pick changes to other branches (based on an issue number, for example) — you need every _parent_ commit :-)
We only cherry-pick when we know exactly what's in there and what's not in there (and is not needed on a different base commit).

I could also tag with an epic number: "Cleanup in the context of this big work".
Just saying.

dcleao:
Changed the commit messages.

test-js/unit/pentaho/lang/Base.Spec.js
```diff
@@ -1322,6 +1385,44 @@ define([
...
+                  Derived.mix({a: 1, b: 2}, null, {exclude: {a: true}});
+
+                  expect(Derived.prototype.a).toBe(undefined);
+                  expect(Derived.prototype.b).toBe(2);
+                });
+
+                it("should ignore the specified properties of the given class specification", function() {
+                  var Derived = Base.extend();
+
+                  Derived.mix({}, {a: 1, b: 2}, {exclude: {a: true}});
```
nantunes: Why a common exclude configuration for instance and class extensions?

test-js/unit/pentaho/lang/Base.Spec.js
```diff
@@ -789,202 +816,206 @@ define([
...
 
-                    expect(persa_cat.color).toBe("bald blackish");
-                  });
+              function expectIt(spec) {
+                var index = 1, aIndex, bIndex, cIndex;
 
-                  it("should be able to call its base method even if there isn't one", function() {
-                    expect(persa_cat.run()).toBeUndefined();
-                  });
+                var Derived1 = Base.extend({
```
nantunes: What happens with:
```javascript
var Derived1 = Base.extend({
  extend_order: ["a", "b", "c", "d"],
  ...
});

var Derived2 = Base.extend({
  extend_order: ["c", "e", "b", "a", "f"],
  ...
});
```
?

package-res/resources/web/pentaho/lang/Base.js
```diff
@@ -35,10 +35,17 @@
...
+ *     that have an own initialization method (specified with the `constructor` property).
  * 14. Added new Class.implementStatic method to allow to specify the class interface after extend.
  *     Is parallel to Class.implement.
  * 15. Any existing static methods are inherited (not only standard Base ones).
+ * 16. Dropped support for the overload Base#extend(name, value).
+ *     The same method now supports only the signature Base#extend(instSpec[, keyArgs]).
+ * 17. Added support to augment the set of per-class, inherited property names that are excluded from
+ *     instance-extension operations.
+ *     Specify `extend_exclude` when sub-classing, e.g. `Base.extend({extend_exclude: {a: true, b: true}});`.
+ * 18. Added support to control, per-class, the properties extension order in instance-extension operations.
```
nantunes: What's the use case for this? Did you run into this when implementing something in particular?

nantunes: Ok, I am reviewing each commit individually. I now saw you are using it in 59abcb7032c8483a9b13011c1566607df1964772.

test-js/unit/pentaho/lang/Base.Spec.js
dcleao: Because it was the case I needed and was simpler to implement.

test-js/unit/pentaho/lang/Base.Spec.js
dcleao: The base order wins. Duplicate elements are only processed in their first occurrence

PR #711, opened 2016-03-28
[BACKLOG-6052] Have specification serialization for foundation classes

dcleao:


dcleao:
Added additional commits. Will merge fixes into corresponding previous commits, once review is over.

dcleao:
I included the changes in https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/646 here. This is out of convenience, as I needed to refactor the Context.Spec.js, to add more tests.
It is now in an even better position to change to any other requirejs testing framework we decide to change to.

nantunes:
Reviewed 89 of 89 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[package-res/resources/web/pentaho/type/complex.js, line 843 \[r1\]](https://reviewable.io:443/reviews/pentaho/pentaho-platform-plugin-common-ui/711#-KE1aLVpGCmDOWHzSoBM:-KE1aLVqg8yO2F2f5YaA:1237842728) ([raw file](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/2777d6b9648b6950b210a7c6138ba1c5a9e926bd/package-res/resources/web/pentaho/type/complex.js#L843)):*
If ```this.count > 0``` both ```any``` and ```spec.props``` will be reset bellow. Is this intended? If so, should this call be moved to an else block?

---

*[package-res/resources/web/pentaho/type/valueHelper.js, line 86 \[r1\]](https://reviewable.io:443/reviews/pentaho/pentaho-platform-plugin-common-ui/711#-KE1dnoopTRJq1O4wlUn:-KE1dnoopTRJq1O4wlUo:962730374) ([raw file](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/2777d6b9648b6950b210a7c6138ba1c5a9e926bd/package-res/resources/web/pentaho/type/valueHelper.js#L86)):*
Would it make sense to call ```valueOf()``` until we get the unwrapped method?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/pentaho/pentaho-platform-plugin-common-ui/711)*

dcleao:
@nantunes 
1. `any` only changes from `false` to `true`, so there's no problem in overwriting the value with `true`. It can already be `true`.
2. `spec.props` is only filled-in in this method. Replacing the value also has no problem.

But I do see that if all properties are inherited there will be no properties in the props array. I added this "inherited" condition later and didn't realize it could become empty. Will change this.

3. There would be no way to know where to stop. I think we could end up in an implementation of the base type or its base, etc.

nantunes:
@dcleao Yeh, my mind was on Base and its ```methodCallsBase``` and ```method.toString == properFunToString``` stuff.

Well, in principle it shouldn't be wrapped twice, anyway.

package-res/resources/web/pentaho/type/_doc/spec/IValueTypeProto.jsdoc
```diff
@@ -0,0 +1,103 @@
+/*!
+ * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
```
carlosrusso: 2016

package-res/resources/web/pentaho/type/_doc/spec/IValueProto.jsdoc
```diff
@@ -0,0 +1,53 @@
+/*!
+ * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
```
carlosrusso: 2016

package-res/resources/web/pentaho/type/_doc/spec/IValueTypeProto.jsdoc
```diff
@@ -0,0 +1,103 @@
...
+ * @memberOf pentaho.type.spec.IValueTypeProto#
+ * @type {?string}
+ */
+
+/**
+ * A reference to the value type's base type, if any.
+ *
+ * When `null`, it has no base type.
+ *
+ * When absent or `undefined`, it defaults to `"complex`.
```
carlosrusso: I suppose you meant `"complex"`

package-res/resources/web/pentaho/type/_doc/spec/IValueTypeProto.jsdoc
```diff
@@ -0,0 +1,103 @@
...
+ * @name base
+ * @memberOf pentaho.type.spec.IValueTypeProto#
+ * @type {?pentaho.type.spec.UTypeReference}
+ */
+
+/**
+ * Indicates if the type is abstract.
+ *
+ * An abstract type cannot be used directly to create a value instance,
+ * and, as such, isn't usually exposed to users in a user interface.
```
carlosrusso: I prefer to avoid contractions: isn't -> it is not

dcleao:
Rewrote history to move fix commits to corresponding main commits again.

dcleao:
Rewrote history to move fix commits to corresponding main commits again.
(unfortunately, github does not show every commit above in the actual order...)

dcleao:
Had already been validated by my colleagues. No one's available now, so I am merging it myself.

PR #714, opened 2016-03-29
[BACKLOG-6739] Prevented calling Object.freeze on null.

carlosrusso:


carlosrusso:
@pentaho/millenniumfalcon Should restore DevCI's happiness.

PR #715, opened 2016-03-30
[BACKLOG-6052] Changed event listeners to better handle key press and a flag to ignore the confirmation window

davidmsantos90:


nantunes:
:lgtom:

PR #718, opened 2016-03-31
[BISERVER-13167] Specifying locale parameter with dialekt in URL works incorrect

IvanNikolaychuk:
What was done:

- We can't depend on URL, when extracting locale, as in the moment when this script is executed url looks like this:  
http://localhost:8080/pentaho/api/repos/pentaho-interactive-reporting/prpti.new?ts=1458641503424, and contains no information about locale.

And there is actually no need to extract url parameter each time, as we have a SESSION_LOCALE for this, that is always up-to-date.
- We make pattern more strict: a valid value of a locale parameter is "xx" (language only, no dialect) or "xx_YY" (language with dialect).  If this pattern is not met, we set locale to be 'en' by default.

-  Before setting locale parameter to dojoConfig object we normalize it according to [dojo requirements] (https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/dojo/pentaho/common/overrides/dojo/_base/config.js#L54).

More detailed description can be found in JIRA comments to this task.

@rfellows , could you please review it?

### DIFF
--- a/package-res/resources/web/dojo/djConfig.js
+++ b/package-res/resources/web/dojo/djConfig.js
@@ -48,13 +48,23 @@
 
 // don't overwrite this if they've set dojoConfig ahead of time
 if(dojoConfig == 'undefined' || dojoConfig == undefined) {
-  var url = (window.location != window.parent.location) ? document.referrer: document.location.href;
   var dojoConfig = {
     disableFlashStorage: true, /* turn off flash storage for client-side caching */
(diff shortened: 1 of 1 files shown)


eliofreitas:
Rejected from automatic build ,authorized users only

IvanNikolaychuk:
@rfellows , could you please review it?

PR #723, opened 2016-04-01
[BISERVER-12918] - Added text input back in cases we dont need an aut…

diogofscmariano:
…ocomplete component. Tests updated accordingly

@tmorgner @MikhailHubanau @pamval please review
@tmorgner i didnt have to check the strict for this use case. What you think?


tmorgner:
"Strict" checking is not needed. All you need to know is whether the parameter is backed by a query or not, and you get that via your " && args.param.list" check. 

If it is a list-parameter you have a selection for the user to choose from, and "autocompletebox" is the correct choice, else the plain textbox is the correct choice. 

So your code is doing exactly what is necessary and I think is good to go.

PR #725, opened 2016-04-04
[BACKLOG-6733] Allow config of local require prior to its first use

nantunes:
* `require.using` receives `config` as second argument
* `config` can either be a function to be called with the cloned require or an object to pass to the `config` method
* the previous `using` method signature is still supported

@dcleao This resolves the issue we were seeing last friday. Please review.


dcleao:
Cool!

PR #734, opened 2016-04-05
[BACKLOG-6733] pentaho/lang/SortedList

nantunes:

### DIFF
--- a/package-res/resources/web/pentaho/lang/SortedList.js
+++ b/package-res/resources/web/pentaho/lang/SortedList.js
@@ -1,0 +1,228 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define([
+  "./List",
+  "./Base"
+], function(List, Base) {
+
+  var baseProto = Base.Array.prototype;
+
+  return List.extend("pentaho.lang.SortedList", /** @lends  pentaho.lang.SortedList# */{
+    /**
+     * @classdesc The `SortedList` class is an abstract base class for ordered arrays.
+     *
+     * ### AMD
+     *
+     * To obtain the constructor of this class,
+     * require the module `"pentaho/lang/SortedList"`.
+     *
+     * ### Remarks
+     *
+     * If an attempt is made to add a element to a specific index of the list
+     * an error is thrown.
+     *
+     * @class
+     * @name SortedList
+     * @memberOf pentaho.lang
+     * @abstract
+     * @extends pentaho.lang.List
+     *
+     * @description Initializes a sorted list instance.
+     *
+     * Note that because a `SortedList` is a sub-class of `Array`,
+     * it cannot be a "newable" constructor function (at least up to ECMAScript version 5).
+     * Instead, instances of `SortedList` are actually initial instances of `Array`
+     * whose prototype is then changed to be that of `SortedList`.
+     *
+     * In other words, `SortedList` is an "initialization" constructor
+     * (see {@link pentaho.lang.ISpecifiable} for more information
+     *  on these concepts).
+     *
+     * Concrete `SortedList` sub-classes should provide a static `to` method
+     * to help in their construction.
+     *
+     * @param {Object}   [keyArgs]         The keyword arguments.
+     * @param {Function} [compareFunction] Specifies a function that defines the sort order.
+     *
+     * These are not used directly by the `SortedList` class
+     * but are passed-through to the methods that handle
+     * the initialization of each list element.
+     */
+    constructor: function(keyArgs, compareFunction) {
+      if (compareFunction) {
+        this.compare = compareFunction;
+      }
+
+      this.base(keyArgs);
+    },
+
+    _compareFunction: function(e1, e2) {
+      if (e1 == null || e2 == null) {
+        if (e1 != null) {
+          return 1;
+        }
+
+        if (e2 != null) {
+          return -1;
+        }
+
+        return 0;
+      }
+
+      var v1 = e1.valueOf();
+      var v2 = e2.valueOf();
+
+      return v1 === v2 ? 0 : (v1 > v2 ? 1 : -1);
+    },
+
+    get compare() {
+      return this._compareFunction;
+    },
+
+    set compare(compareFunction) {
+      this._compareFunction = compareFunction;
+    },
+
+    sort: function(compareFunction) {
+      baseProto.sort.call(this, compareFunction || this._compareFunction);
+    },
+
+    /**
+     * Performs a binary search on the sorted list.
+     *
+     * @param {*} elem The item to search for.
+     *
+     * @return {Number} The index of the element;
+     *                  when not found returns the negative
+     *                  index of the closest element.
+     */
+    search: function(elem) {
+      var left = 0;
+      var right = this.length - 1;
+
+      while (left <= right) {
+        var i = (left + right) / 2 | 0;
+
+        var comparison = this.compare(this[i], elem);
+
+        if (comparison < 0) {
+          left = i + 1;
+        } else if (comparison > 0) {
+          right = i - 1;
+        } else {
+          return i;
+        }
+      }
+
+      return ~right;
+    },
+
+    copyWithin: function() {
+      throw new Error("Can't copy within a sorted list.");
+    },
+
+    fill: function() {
+      throw new Error("Can't fill a sorted list.");
+    },
+
+    reverse: function() {
+      throw new Error("Can't reverse a sorted list.");
+    },
+
+    unshift: function() {
+      throw new Error("Can't do a indexed insert in a sorted list.");
+    },
+
+    insert: function() {
+      throw new Error("Can't do a indexed insert in a sorted list.");
+    },
+
+    replace: function() {
+      throw new Error("Can't do a indexed replace in a sorted list.");
+    },
+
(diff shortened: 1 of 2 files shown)


package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+     * to help in their construction.
+     *
+     * @param {Object}   [keyArgs]         The keyword arguments.
+     * @param {Function} [compareFunction] Specifies a function that defines the sort order.
+     *
+     * These are not used directly by the `SortedList` class
+     * but are passed-through to the methods that handle
+     * the initialization of each list element.
+     */
+    constructor: function(keyArgs, compareFunction) {
```
dcleao: Why not use keyArgs to specify the compareFunction?

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+        return 0;
+      }
+
+      var v1 = e1.valueOf();
+      var v2 = e2.valueOf();
+
+      return v1 === v2 ? 0 : (v1 > v2 ? 1 : -1);
+    },
+
+    get compare() {
```
dcleao: Would prefer naming this as `comparer`, but I see the idea of also directly calling `compare(a, b)`

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+      return this._compareFunction;
+    },
+
+    set compare(compareFunction) {
+      this._compareFunction = compareFunction;
+    },
+
+    sort: function(compareFunction) {
```
dcleao: Ain't accepting a custom sort function as much a violation as allowing to insert at arbitrary indexes? A subsequent call to search would fail.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+      baseProto.sort.call(this, compareFunction || this._compareFunction);
```
dcleao: Also trying to understand the use of `baseProto` instead of `this.base`. I admit it is faster this way.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+     * @return {Number} The index of the element;
+     *                  when not found returns the negative
+     *                  index of the closest element.
+     */
+    search: function(elem) {
+      var left = 0;
+      var right = this.length - 1;
+
+      while (left <= right) {
+        var i = (left + right) / 2 | 0;
```
dcleao: Please add a comment explaining that the `x | 0` is equivalent to `Math.floor(x)`.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+        if (comparison < 0) {
+          left = i + 1;
+        } else if (comparison > 0) {
+          right = i - 1;
+        } else {
+          return i;
+        }
+      }
+
+      return ~right;
```
dcleao: Please add a comment saying this performs a two's complement operation.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+    sort: function(compareFunction) {
+      baseProto.sort.call(this, compareFunction || this._compareFunction);
+    },
+
+    /**
+     * Performs a binary search on the sorted list.
+     *
+     * @param {*} elem The item to search for.
+     *
+     * @return {Number} The index of the element;
```
dcleao: Please replace `Number` by `number`. Also,  shouldn't the description say that the result is the two-complement index of where the element should be inserted to be in order (which is not the same as the closest element?).
When not found, an insertion operation should insert the element at `~result`, and not `-result`, AFAIK.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+      }
+
+      baseProto.splice.apply(this, arguments);
+    },
+
+    _insertInOrder: function(elem, keyArgs) {
+      var elem2 = this._adding(elem, null, keyArgs);
+
+      if (elem2 !== undefined) {
+        var index = Math.abs(this.search(elem2));
```
dcleao: As said above, not sure about this abs here.
```js
var index = this.search(elem2);
if(index < 0) index = ~index;
baseProto.splice....
```

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -0,0 +1,228 @@
...
+      var v2 = e2.valueOf();
+
+      return v1 === v2 ? 0 : (v1 > v2 ? 1 : -1);
+    },
+
+    get compare() {
+      return this._compareFunction;
+    },
+
+    set compare(compareFunction) {
```
dcleao: Taking care for the initialization step, setting should sort the array.
Setting to null should reset the local value, thus inheriting the prototype function again?

brosander:
http://wingman.pentaho.com:8080/job/wingman/58/

brosander:

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: You're right, missed the obvious.

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: Because it doesn't make sense for the elements type.
> These are not used directly by the `SortedList` class but are passed-through to the methods that handle the initialization of each list element.

(also noticed the above sentence belongs near the `keyArgs`... as is it seems to apply to both parameters)

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: Was mimicking the `pentaho.lang.List` without noticing it is using it to call **other** prototype functions other than the one available at `this.base`. Will change for consistency (and also in `splice`).

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: The ~ complement operation (as a side effect, I admit) when dealing with integers results in `~x = (-1 * x) - 1`.

So lets recapitulate:

* When not found, we get the _negative index minus 1_ of the _last smaller element_. `Math.abs` it and we get the index to insert our element **after** that one.
 * In an empty list, or in a list with only _bigger elements_, it will be equal to `0`. In a list with only _smaller elements_ it will be equal to `.length`. So it makes sense.
* When found we get the index of the first _equal element_. Inserting at it adds the new element **before** that one.

Maybe this second case is the one that can prevent
> Within a type configuration, if all other rule ordering criteria are equal, the later configuration rules override the former configuration rules.

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: Good catch!

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: As for the wording, agreed... Just don't know how to say it better...

package-res/resources/web/pentaho/lang/SortedList.js
dcleao: The base `List` class did not find any direct use for `keyArgs`. However, the options bag was meant to satisfy the construction of both the `List` and its `Element`. At least, that was my original intent.

nantunes: Yes, but how do you see it working? Should the List remove the `compareFunction` key from the `keyArgs` passed on to the `Element`? Or keep it? What if the `Element` also needs an argument called `compareFunction` (a `SortedList` of `SortedList`s?).

nantunes:
@pentaho/millenniumfalcon New commit incorporating some of the reviews.

dcleao:
You're right on the insert sort indexes issue.

package-res/resources/web/pentaho/lang/SortedList.js
```diff
@@ -93,11 +95,24 @@ define([
     },
 
     set compare(compareFunction) {
-      this._compareFunction = compareFunction;
+      var hasOwn = O.hasOwn(this, '_compareFunction');
+      if (hasOwn && this._compareFunction !== compareFunction || !hasOwn && compareFunction != null) {
+        if (compareFunction == null) {
+          O.delete(this, '_compareFunction');
```
dcleao: I don't think that there is a reason to use the `O.delete` method here. Actually, I think it's not being used anymore and we should probably remove it.

dcleao:
👍

nantunes:
@dcleao My comment https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/734#discussion_r58684638 was hidden because of the outdated diff. Have you seen it?

package-res/resources/web/pentaho/lang/SortedList.js
nantunes: Yeh, but that calling is internal and the setter name is public... So will change to your suggestion.

package-res/resources/web/pentaho/lang/SortedList.js
dcleao: Yes, I see it in a kind of messy way... Reserving the name on the SortedList constructor implies we will not be able to specify a same named argument for an Element keyArg.
Ideally we would "block" the List argument from leaking to the Elements, but that implies either:
1. creating an inherited object and locally setting the property to `undefined`, kind of shadowing it 2. deleting the property, mutating the given keyArgs object
3. creating a copy of the keyArgs without the option

All are costly solutions, so I usually just let these pass-through and live with the imposed argument name restriction imposed on Elements...

There's also the option of having a separate keyArgs for each (or one inside the other), but it appears to be over-engineering. Dunno.

nantunes:
@pentaho/millenniumfalcon Please check if it is now good to merge.

dcleao:
Don't take me wrong. I don't see an advantage in dissociating the argument name `compareFunction` from the property name `comparer`. I do see the point in choosing an improbable name, but think it's way unattractive for a public name.
Apart from this, all looks good.

nantunes:
@dcleao Done.

PR #746, opened 2016-04-11
[BACKLOG-6585] - adding more options to the FormatCombo. also added c…

rfellows:
…ategory for future usage to further filter dow the options displayed.


rfellows:
@DFieldFL please review

PR #750, opened 2016-04-12
[BACKLOG-7199] Fatal error occurs in case async report generation and…

EgorZhuk:
… empty page number text box

@tmorgner here we have only one commit, all other are covered by OrLando CaLisbon team changes


brosander:

PR #751, opened 2016-04-12
[CLEANUP] Transformed absolute AMD paths to relative, removed referen…

carlosrusso:
…ces to unused dependencies

@pentaho/millenniumfalcon please review. I've been piling these cleanups over time.

### DIFF
--- a/package-res/resources/web/pentaho/data/_filter/_Element.js
+++ b/package-res/resources/web/pentaho/data/_filter/_Element.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 define([
-  "pentaho/lang/Base"
+  "../../lang/Base"
 ], function(Base) {
   "use strict";
 
--- a/package-res/resources/web/pentaho/util/object.js
+++ b/package-res/resources/web/pentaho/util/object.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-define(["pentaho/util/has"], function(has) {
+define(["./has"], function(has) {
   "use strict";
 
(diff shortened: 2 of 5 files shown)


brosander:

package-res/resources/web/pentaho/data/_filter/_Element.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 define([
-  "pentaho/lang/Base"
+  "../../lang/Base"
```
nantunes: I understand this is consistent with other cases in the code base, but I question if it should... Why not use the absolute ID?

package-res/resources/web/pentaho/util/object.js
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-define(["pentaho/util/has"], function(has) {
+define(["./has"], function(has) {
```
nantunes: The same above question.

package-res/resources/web/pentaho/data/_filter/_Element.js
carlosrusso: I don't think that the absolute ID is wrong.
My reasoning:
1. It becomes consistent with the rest of the code base (as you pointed out)
2. My IDE resolves the paths with minimal fussing of the configuration (yes, it's selfish)

nantunes: Ok. I think this can be further discussed, but as it is consistent with the current code, will merge.

PR #752, opened 2016-04-13
[CLEANUP] change for run js buld and test for windows

SergeyTravin:


SergeyTravin:
@lgrill-pentaho @mbatchelor Could you please review this PR. 

It contains changes in subfloor which allow to run build js under windows machine without changes the system Path and additional note for karma. Under windows, we should use "npm install karma --save-dev" to create project only copy karma and successfully run the js build. Thanks!

brosander:

lgrill-pentaho:
This work needs to be reviewed by ENGOPS and at least one Lead developer. 
Also needs to be evaluated with mutli-platform test builds. 
This pull request creates un-scoped work for all parties.
Well intentioned and probably a good idea. 
Please close this request and create a dev tech debt or ENGOPS ticket to track this work.

PR #757, opened 2016-04-26
[BACKLOG-7613] When changing selection mode it is not possible to can…

davidmsantos90:
…cel the selection in a visualization


brosander:

dcleao:
👍

PR #760, opened 2016-04-27
[BACKLOG-7356] - Added ability to handle array parameters properly ev…

diogofscmariano:
…en when types are not correct. Tests updated accordingly


brosander:

krivera-pentaho:
Is it ok that this build failed wingman test?

pamval:
yup, wingman is having some issue with building common-ui. @diogofscmariano is looking into that.
in the meantime, review manually, please.

PR #762, opened 2016-04-27
[BACKLOG-7614] Serialize a visualization to JSON

dcleao:
NOTE: Documentation will be reviewed along with all the remaining metadata docs.

### DIFF
--- a/test-js/unit/pentaho/type/simple.serialization.Spec.js
+++ b/test-js/unit/pentaho/type/simple.serialization.Spec.js
@@ -204,68 +204,113 @@ define([
     });
   }); // pentaho.type.Simple
 
-  var simplerTypes = [
-    {name: "boolean",  value: true},
-    {name: "number",   value: 10},
-    {name: "string",   value: "hello"},
-    {name: "object",   value: {foo: "bar"}},
-    {name: "function", value: function() {}},
-    {name: "date",     value: new Date()}
-  ];
+  //region Other Simple Types Test Helpers
 
-  simplerTypes.forEach(function(simpleSpec) {
-    describe("pentaho.type." + simpleSpec.name, function() {
-      var SimpleClass = context.get("pentaho/type/" + simpleSpec.name);
+  function testSimpleCommon(SimpleClass, primitiveValue) {
 
-      var value;
-      beforeEach(function() {
-        value = new SimpleClass(simpleSpec);
-      });
+    it("should output a cell with the '_' inline type reference when includeType: true", function() {
+      var value = new SimpleClass({v: primitiveValue});
+      var spec = value.toSpec({includeType: true});
 
-      describe("values", function() {
-        describe("primitive format (omitFormatted: true)", function() {
-          it("should return the primitive value when it is not a plain object or wrap it otherwise", function() {
-            var spec = value.toSpec({omitFormatted: true, includeType: false});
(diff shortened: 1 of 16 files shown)


brosander:

test-js/unit/pentaho/type/simple.serialization.Spec.js
```diff
@@ -158,10 +224,13 @@ define([
 
       describe("values", function() {
         describe("primitive format (omitFormatted: true)", function() {
-          it("should return the primitive value", function() {
+          it("should return the primitive value when it is not a plain object or wrap it otherwise", function() {
             var spec = value.toSpec({omitFormatted: true, includeType: false});
-
-            expect(spec).toBe(simpleSpec.value);
+            if(simpleSpec.value instanceof Object && simpleSpec.value.constructor === Object) {
```
nantunes: Multiple paths in a unit test should be avoided.
I understand this is running in the `simplerTypes.forEach` loop, but who to make clear to whoever is reading the code which of the above should or shouldn't be wrapped?

nantunes:
Maybe you could add one test (or more) asserting that `toJSON` is being called when `JSON.stringify` is used.

dcleao:
Yes, even though `toJSON` being called by `JSON.stringify` is standard, it's good to be sure we did nothing wrong.

test-js/unit/pentaho/type/simple.serialization.Spec.js
dcleao: 1. Generally, agree.
2. In a way, the test title says, almost directly, which cases do what: "when the primitive value is a plain object, it returns the wrapped value", and the case `{name: "object",   value: {foo: "bar"}}` is the only case that has a primitive value which is a plain object.
3. A solution could be to add a new property to each case info object, like `isValuePlainObject`. This would allow placing the branching outside the `it` and, instead, generate one of two distinct `it`s.
4. Another solution would be to undo/refactor the forEach construction, possibly splitting it into several groups of assertions, so that the plain object case could be handled specially
5. In yet another solution, it would be left as is (because sometimes the cure is worse than the disease!).

Suggestions anyone?

dcleao:
Just added two new tests for testing toJSON is actually called by JSON.stringify, in a new commit. Needs squash!

test-js/unit/pentaho/type/simple.serialization.Spec.js
nantunes: 🔢 I would go for option 4 (extract the singular behaviour of the plain object) or 5 (maybe with a simple comment somewhere).

nantunes:
@dcleao Are you dealing with circular references?
I know the built-in JSON serialiser also doesn't, but I just run into a `TypeError: Converting circular structure to JSON` when doing a `JSON.stringify(model.toSpec());`, so remembered asking.

dcleao:
Refactored the simple types' tests according to proposal 4. Need squash.

dcleao:
@nantunes that's because the model currently contains objects with circular references, like the `DataTable`. Even if you'd use `JSON.stringify(model)`, which calls `toJSON` instead of `toSpec`, you'd probably have the same problem...
This might be the time we need to create a Simple type subclass for holding the data table, and, possibly, one for the data filters as well...

PR #770, opened 2016-05-03
[BACKLOG-7551] Check the content of a value before setting it for an update change

davidmsantos90:
This is to prevent simple values from being updated, when in reality their content (value, formatted) didn't change.

To escape the need to use 'instanceof' and because we still didn't tackle the Update issue, for now I implemented a default method in element.js that always returns `false`.

Also noticed that a complex value most likely will never be updated with our current implementation of the list algorithm, because when two complex values have the same `key`, even though their content might have changed the object is the same, making the test `existing !== elem` return `true` instead of `false`.

### DIFF
--- a/package-res/resources/web/pentaho/type/simple.js
+++ b/package-res/resources/web/pentaho/type/simple.js
@@ -194,6 +194,24 @@ define([
         return this._value.toString();
       },
 
+      /**
+       * Determines if a given value, of the same type, represents the same entity with the same content.
+       *
+       * The given value **must** be of the same concrete type (or the result is undefined).
+       *
+       * If two values are equal, they must have an equal [key]{@link pentaho.type.Simple#key},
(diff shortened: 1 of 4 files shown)


package-res/resources/web/pentaho/type/simple.js
```diff
@@ -194,6 +194,24 @@ define([
...
+       * [value]{@link pentaho.type.Simple#value}, [formatted value]{@link pentaho.type.Simple#formatted}.
+       *
+       * @param {!pentaho.type.Simple} other - A simple value to test for equality.
+       * @return {boolean} `true` if the given simple value is equal to this one, `false`, otherwise.
+       */
+      equalsContent: function(other) {
+        if(!this.equals(other)) return false;
+
+        // TODO: generic metadata
+        return this._value === other._value && this._formatted === other._formatted;
```
dcleao: `_value` comparison is unnecessary no?

dcleao:
👍

PR #771, opened 2016-05-03
[BACKLOG-7614] Added docs on native functions not being serialized.

dcleao:


dcleao:
This one is a good one (testing slack).

nantunes:
@dcleao I think it will only work on new stuff, not with comments.

PR #780, opened 2016-06-10
[BISERVER-13154] Uncaught TypeError: Cannot read property decimalForm…

ivanpogodin:

### DIFF
--- a/test-js/unit/dojo/pentaho/DojoNumberSpec.js
+++ b/test-js/unit/dojo/pentaho/DojoNumberSpec.js
@@ -1,0 +1,45 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+define( ["dojo/number"], function( DojoNumber ) {
+
+  describe( "Dojo i18n tests", function() {
+
+    it( "Number.parse(..) should accept default locale", function() {
+      expect( DojoNumber.parse( "1234", {} ) ).toEqual( 1234 );
+    } );
+
+    it( "Dojo Number.parse(..) should accept correct locale", function() {
+      expect( DojoNumber.parse( "1234", {"locale" : "en-us"} ) ).toEqual( 1234 );
+    } );
+    
+    // This is a known dojo issue
+    // See BISERVER-13154
+    it( "Dojo Number.parse(..) should safely accept even incorrect locale", function() {
+      //expect( DojoNumber.parse( "1234", {"locale" : "en_us"} ) ).toEqual( 1234 );
+      expect( function(){ DojoNumber.parse( "1234", {"locale" : "en_us"} ) } ).toThrowError( TypeError );
+    } );
+
+    // This is a known dojo issue
+    // See BISERVER-13154
+    it( "Dojo Number.parse(..) should safely accept even abracadabra locale", function() {
+      //expect( DojoNumber.parse( "1234", {"locale" : "abracadabra"} ) ).toEqual( 1234 );
+      expect( function(){ DojoNumber.parse( "1234", {"locale" : "abracadabra"} ) } ).toThrowError( TypeError );
+    } );
+
+  } );
+
+} );
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -632,6 +632,13 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
          * @returns {Object} parameters The parameters name|value pair assigned to the dashboard instance
          */
         getParameterValues: function () {
+          function parseNumber(val){
+            try{
+              return DojoNumber.parse(val, { locale : Util.normalizeDojoLocale(SESSION_LOCALE) });
+            } catch(e) {
+              return DojoNumber.parse(val, { locale : "en" });
+            }
+          }
           var params = {};
           this.getParamDefn().mapParameters(function (param) {
             var value = this.getParameterValue(this.getParameterName(param));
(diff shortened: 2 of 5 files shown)


ivanpogodin:
@rfellows , @carlosrusso , @tmorgner , could you review it, please?
FYI
There is an additional fix of a probable similar problem (see https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/781). Should it be included?

rfellows:
This looks good. However, if you think that #781 should be included i'd like to see that function moved into a utility somewhere so it's not duplicated in all the places that need it. seems like a useful little bit of code.

tmorgner:
I seriously doubt that your fix solves the problem. This is a more complex "toLowerCase" method. 

However, that would not matter at all, as DojoNumber.format(..) calls i18.normalizeLocale(..) which also does the toLowerCase() call. The "decimalFormat" property is only used in dojo/number.js#format and tbh at no point should that code throw an error unless you cannot load the bundle.

So could you check whether Dojo actually loads the locale-bundle correctly? In theory it should not even fail if the locale is completely wrong (ie you use "gobbeldigock" as locale) as Dojo is supposed to fall back to a default-locale. So that needs proper investigation, maybe in form of a unit-test to see that you can replicate the error more reliably and to prove that the error is gone.

ivanpogodin:
@tmorgner ,
Unfortunately, i18.normalizeLocale(..) does not correct underscores.
The toDojoCompatibleLocale(..)  replaces underscore with hyphen (as supposed in dojo/i18n.js).
May be, i18.normalizeLocale(..) should be called rather then toLowerCase(..).

There is a similar fix in "djConfig.js" (https://github.com/pentaho/pentaho-platform-plugin-common-ui/commit/4c71f8f13e24c527e428b0e3c430fe1d76b38d22).

tmorgner:
But even with an totally invalid locale, the code should not crash. According to the Dojo documentation, If dojo cannot find a locale, it should go back to the master bundle. (This is mentioned just in one place in the Dojo documentation: https://dojotoolkit.org/documentation/tutorials/1.10/i18n/ )

Dojo loads locales via requirejs and therefore the locale string here is merely a lookup key. If I understand the dojo resource-bundle loading correctly, the content of that text should not matter as long as dojo can find the file in the correct directory.

So why does the formatting-call fail if the locale content is supposed to be ignorable? 

In the other pull request you mention that you have not been able to replicate the bug. Can you try to trigger the bug in some sort of scenario so that we can actually get to the true cause of that bug?

ivanpogodin:
@rfellows , @carlosrusso , @tmorgner , could you review it, please?
https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/780/commits/b5c3550424cf91b3f26b849abbaec26ef9c7c71b
Shared code is moved to util.js

https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/780/commits/d95e6c6c4ce43331108b9b1b70f4b2cc48ea1ac7
There are failing tests for dojo/number.
They show the cause of this issue.

tmorgner:
wow, so we are talking about a Dojo bug here. In theory, Dojo should try out all locales (see the very long comment starting at dojo/i18n.js line 120 on how this is supposed to work. 

If I run the test, I cannot get the "en-us" (should use correct locale) to pass. There is no "cldr/nls/en-us" directory, so the direct lookup will fail. If I change the test to "en-gb", it passes. 

I added some additional tests for other minority languages (de-de, de-ch) and they predictably fail too. At this point I feel just replacing bundle names and praying wont do the trick. 

From digging a bit into the source code I see that Dojo tries to load the bundle-modules via RequireJS and tries to load them in a synchronous way. (see i18n.js#doLoad(..) and i18n.js#getLocalization(..). 

AFAIK the RequireJS function "require" will return immediately, so the code in "doLoad" will not work as intended.  At this point I am at the end of my knowledge and have no additional insights to offer other than that this code can break for a lot of users very quickly. 

@rfellows, @carlosrusso - how would you teach Dojo to handle locales gracefully?

rfellows:
I don't think there is much we can do from the outside. This really seems like a dojo issue that we have to work around.

dcleao:
In the `getLocalization` function, either `syncRequire` or `require` are passed to `doLoad`, depending on whether the request is cross-domain (or something like that). Anyway, the function `syncRequire` is provided by dojo and tries its best to perform a sync request.

ivanpogodin:
A test for the util function is added, as a result the function is improved.
The additional dojo/number test is left in a special commit.

tmorgner:
@dcleao It may have an option for trying sync calls, but then the tests should not fail. 

The tests will also fail for unknown locales like "de-ch" or "pt-br". There are quite a few people in these language zones who will see the error. 

You only get a "undefined" as bundle if the callback function that assigns the loaded bundle to the "load(..)" function's return value is never called. 

At the moment the code is still a "hope and pray" as demonstrated by the failing test. So at minimum I would actually check whether dojo can load the resource bundle:

       // from dojo's number.parse(..)
	var locale = i18n.normalizeLocale(options.locale),
		bundle = i18n.getLocalization("dojo.cldr", "number", locale);

and test the bundle for "undefined", or wrap each and every DojoNumber.parse into a try-catch that retries the parsing without a locale (which then hopefully does not fail) so that this Dojo error does not break other functionality.

ivanpogodin:
@tmorgner , yes, unknown locales should be processed safely.

ivanpogodin:
@tmorgner , 
I've avoided using overrides (package-res\resources\web\dojo\pentaho\common\overrides\dojo\).
Just try-catch instead of direct call of dojo. Is this an appropriate solution?
The util.normalizeDojoLocale() is not neccessary now, though it makes program a bit smarter (underscored locales are not replaced with "en").

ivanpogodin:
@rfellows , @carlosrusso , @tmorgner , @dcleao , could you review it, please?
By the way, I commit changes and Wingman shows code coverage changed (increased or decreased - both possible) though java code is unchanged. It looks like there are not determinate tests.

ivanpogodin:
Code coverage can be shown changed even without any code modified.

pamval:
@ivanpogodin can you squash the commits, please?

ivanpogodin:
@pamval 
squashed

test-js/unit/dojo/pentaho/DojoNumberSpec.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it( "Dojo Number.parse(..) should accept correct locale", function() {
+      expect( DojoNumber.parse( "1234", {"locale" : "en-us"} ) ).toEqual( 1234 );
+    } );
+    
+    //Warning!
+    // This test is failing.
+    // See BISERVER-13154
+    it( "Dojo Number.parse(..) should safely accept even incorrect locale", function() {
+      expect( DojoNumber.parse( "1234", {"locale" : "en_us"} ) ).toEqual( 1234 );
```
tmorgner: Given that this test will be failing (and we expect it to), can you change the assert to "expect(..).toThrow()".

This way you both state that this code is expected to fail with the current Dojo code, and your test no longer reports an failing unit-test and thus does not kill the strict build process (the ones that actually fail on test errors).

It might be also wise to update the comment to say: "This test is failing as dojo does not correctly resolve resource-bundles." This simple note saves later developers from having to dig though an ancient JIRA case just to understand why this test is here.

ivanpogodin:
@tmorgner ,
test-js/unit/dojo/pentaho/DojoNumberSpec.js is changed to expect error rather than fail.

ivanpogodin:
@rfellows , @carlosrusso , @tmorgner , could you review it, please?

pamval:
@tmorgner, can you review, please?

tmorgner:
That looks good now. Lets merge.

PR #783, opened 2016-06-14
[BACKLOG-8624] - Make Localization Service available by default in pr…

plagoa:
…oducts (Add to Feature)

- Added new AMD module to access OSGI localization web service
- Changed common-ui require config

### DIFF
--- a/package-res/resources/web/pentaho/i18n.js
+++ b/package-res/resources/web/pentaho/i18n.js
@@ -42,10 +42,10 @@ define(["./util/MessageBundle", "json"], function(MessageBundle) {
   };
 
   /**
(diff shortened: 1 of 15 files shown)


plagoa:
@pentaho/kashyyyk please review

nantunes:
Why another plugin?

dcleao:
Not being aware of all of the details behind the choice of creating a new amd loader plugin, I do want to reinforce that I don't see an impediment to using the existing i18n plugin and to switching between the two environments. One of the purposes for creating the initial plugin was to abstract the consumer from osgi/non-osgi differences...

On another front, on the actual implementation, is there any reason to use a synchronous request call?

PR #792, opened 2016-06-21
[PPP-3514] Use of vulnerable components - jquery - multiple versions …

mchen-len-son:
…1.x prior to 1.12

- updated jquery version to 1.12.4


pamval:
@pentaho-nbaker @rfellows please take a look

mchen-len-son:
@pentaho-nbaker @rfellows can you please take a look?

PR #797, opened 2016-07-04
[SP-2743] Backport of BISERVER-13154 - Uncaught TypeError: Cannot rea…

ivanpogodin:
…dproperty decimalFormat of undefined using PRPT with Number Parameter (6.1Suite)

  pick [BISERVER-13154] Uncaught TypeError: Cannot read property decimalFormat of undefined using PRPT with Number Parameter


mchen-len-son:
There are still test failures due to non-deterministic tests.
Please cherry-pick #701 as well

PR #800, opened 2016-07-08
[BISERVER-13158] Cascading parameters with empty default values and V…

mchen-len-son:
…alidate Values selected causes the report to never render

- added check for dependent strict parameter - it should be created even wothout values

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -92,13 +92,6 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
        * @private
        */
       function _createWidgetForParameter(param) {
-        if (param.strict && param.values.length === 0) {
-          // if the parameter is strict but we have no valid choices for it, it is impossible for the user to give it a
-          // value, so we will hide this parameter it is highly likely that the parameter is driven by another parameter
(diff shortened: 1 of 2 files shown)


mchen-len-son:
Looks like those failing integration tests a non-deterministic ones.
They are all passing locally.

pamval:
@annapopova1 @krivera-pentaho please take a look at this pull request

annapopova1:
Looks good but probably make sense to add some tests to ParameterXmlParserSpec.js also.

tmorgner:
Be aware that the dependency-information you rely on in this fix is part of a new feature (BACKLOG-7980) that does not exist in older versions of the BI-Server and (at this point) is fresh and unvalidated code (the commit for it is just 11 days old, and the story is in active development).

I would say: Cut out the dependency-analysis stuff and go back to the basics, so that you dont have to rely on experimental code.

If there is a parameter defined, and the parameter is not explicitly marked as hidden, there should always be a widget for it. It does not matter that the combobox might be empty - if that is the state of the system, show it as it is, let the report-developer solve it. 

I bet with you if you cut out the if-statement "optimization" in "createWidgetForParameter" the code will behave exactly as it should.It will find a widget when it looks for one, and will omit the parameter from the value-set that is sent back to the server (as there is no selection). Then the server side validation routine can deal with it.

mchen-len-son:
@tmorgner @pamval @diogofscmariano 
guys, I've updated the PR
please, take a look

PR #803, opened 2016-07-15
[PIR-1291] A calendar isnt displayed in Filter pop-up window in PIR r…

mchen-len-son:
…eports

- removed explicit z-index setting


mchen-len-son:
This is a partial revert of #796.

Removing explicit x-index setting will return us to the issue at [BACKLOG-5799 comment (second one)](http://jira.pentaho.com/browse/BACKLOG-5799?focusedCommentId=250174&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-250174).
However, this issue was not intended to be fixed when backporting. Only the 'width problem' was the one we were suffering from.

@pamval @rfellows @tmorgner please take a look at the changes

This is a blocker for 6.1 patch testing
Thanks.

PR #806, opened 2016-07-15
[PIR-1291] A calendar isnt displayed in Filter pop-up window in PIR r…

mchen-len-son:
…eports
- increased z-index for dijit popups

### DIFF
--- a/package-res/resources/web/dojo/dijit/themes/dijit.css
+++ b/package-res/resources/web/dojo/dijit/themes/dijit.css
@@ -121,6 +121,7 @@ TABLE.dijitSelect {
   border: 0;
   padding: 0;
   -webkit-overflow-scrolling: touch;
+  z-index: 2000 !important;
 }
 
 .dijitPositionOnly {


mchen-len-son:
Dijit popups are overlapped by pentaho.common.Dialog after #693.

@annapopova1 @tmorgner @rfellows @pamval please review

lucboudreau:
Please see if you can fix these tests while you're working on this project. Thanks!

mchen-len-son:
After a quick look, I don't see what exactly can cause these test failures.
They don't fail locally.

All I can think of, considering the failures I see, is that the HSQLDB is still initializing at the moment of these tests run.

> `assertTrue( "wrong column name", json.indexOf( "BC_PRODUCTS_PRODUCTLINE" ) != -1 );`
> `assertTrue( "wrong column type", json.indexOf( "\"string\"" ) != -1 );`
> `assertTrue( "wrong value: " + json, json.indexOf( "Classic Cars" ) != -1 );`

See? The first two checks passed, but the last one failed. Looks like the data was incomplete at that moment.

But, still, this is only a guess.

pamval:
is there any other way to fix this issue? Using !important should be avoided at all costs, since there's no way to override or customize this style.

mchen-len-son:
It's not working without !important.

I'll take a look for another ways.

pentaho-lmartins:
Closing PR since JIRA issue is closed.

PR #809, opened 2016-07-22
[BACKLOG-7980] Provide better parameter handling to avoid multiple po…

Aliaksandr-Kastenka:
…sts to the server for revalidations
Goes with https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/400


Aliaksandr-Kastenka:
@tmorgner 
Please review.

PR #812, opened 2016-07-27
[SP-2758][ANALYZER-3289] Issue of reverse gradient color in GeoMap wh…

mchen-len-son:
…en using csv as a datasource

- fixed geo points gradient coloring
- added unit test


mchen-len-son:
@pamval @carlosrusso please review

those failing tests are non-deterministic ones. they fail from time to time, and don't fail locally at all

mchen-len-son:
updated license header

PR #814, opened 2016-07-27
[BACKLOG-9334] - PIR: Report generating is not cancelled if changing …

dzmitry-hryhoryeu:
[BACKLOG-9334] - PIR: Report generating is not cancelled if changing templates by double click.

@tmorgner 
Please review.

### DIFF
--- a/package-res/resources/web/dojo/pentaho/common/TemplatePicker.js
+++ b/package-res/resources/web/dojo/pentaho/common/TemplatePicker.js
@@ -83,6 +83,17 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_Templated", "dojo/on"
             },
 
             imgClick: function (event) {
+              // BACKLOG-9334 - prevent a double-click
+              var me = this;
+              if (!me.clicks) me.clicks = 0;
+              ++me.clicks;
+              if (me.clicks > 1) {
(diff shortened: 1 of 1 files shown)


dzmitry-hryhoryeu:

annapopova1:
@dzmitry-hryhoryeu Could you recheck your fix on IE 11?
I tried check it with my Windows 7 and IE 11 and looks like event.detail returns always 0.

package-res/resources/web/dojo/pentaho/common/TemplatePicker.js
```diff
@@ -83,6 +83,9 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_Templated", "dojo/on"
             },
 
             imgClick: function (event) {
+              if(event.detail > 1){
```
annapopova1: Looks like this line does not match existing code style (even within this single function) - spaces after 'if' and before '{'.

dzmitry-hryhoryeu: @annapopova1 Yes, unfortunately, this solution does not work for the IE. I modified fix and changed styles.

annapopova1:
Looks good to me

PR #815, opened 2016-08-01
[BACKLOG-9339] - 'Error parsing parameter information' when inputting…

Aliaksandr-Kastenka:
… invalid data in the text box parameter

Additional check added in order to avoid exception at this point: paramSelectedValue.toUpperCase().

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -350,13 +350,20 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
           } else {
             switch (paramType) {
               case "java.lang.String": // Used upper case to eliminate UPPER() post-process formula influence on the strings comparison
-                return paramValue.toUpperCase() != paramSelectedValue.toUpperCase();
+                if(!_.isArray(paramSelectedValue)) {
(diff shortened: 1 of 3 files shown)


Aliaksandr-Kastenka:
@pamval 
Please review.

pamval:
@annapopova1 @afrjorge @krivera-pentaho please review this change

annapopova1:
Looks like this error can be happened not only for "java.lang.String" type because as inputs we have paramValue as string and paramSelectedValue as empty array. And after that we try fn toUpperCase for array. Probably for some cases we can have not empty array and as a result potential issue. Another potential case when we have paramSelectedValue as empty string and probably wrong comparison. 
IMO '_areParamsDifferent' function should have smarter logic to compare different types.
In addition will be better to have unit tests to cover possible changes.

Aliaksandr-Kastenka:
@annapopova1 
I checked other possible types of paramSelectedValue(Date and Number) and as far as I see everything works as expected. Added js-test to cover this change.

Aliaksandr-Kastenka:
@pamval 
Please review.

pamval:
@annapopova1 please review this pull request

annapopova1:
Sorry for so late response.
Now code looks good. But I can't say anything about ParameterValidator because I have not had experience of working with it.

PR #819, opened 2016-08-08
[CLEANUP] Linting Pentaho's Client Metadata Model and VizAPI with eslint

dcleao:
Played a bit with eslint. Configured it the best I could to match the linting and formatting rules we presently use in the Client metadata model and VizAPI (generally, the pentaho folder/namespace).

Linted a bit, as a consequence.


package-res/resources/web/pentaho/data/CellTuple.js
```diff
@@ -26,25 +26,29 @@ define([
...
-    //region IWithKey implementation
+    // region IWithKey implementation
     keyName:  "key",
 
     _key: null,
 
     get key() {
       var key = this._key;
-      if(key == null) key = this._key = this.map(function(cell) { return cell.key; }).join("~");
+      if(key == null)
```
nantunes: I would always enforce brackets. But if not, this is now much more dangerous than being in a single line.

dcleao: There's a rule about the maximum number of statements in a single line. I fixed that one, but should have added parenses as well. Anyway, I wouldn't go as far as qualifying it as "much more _dangerous_"...
Anyway, my goal with this PR is not to discuss the current rules — which would take time we don't currently have — but simply to configure the current rules in an ESLint file.

I deactivated the parens rule; currently, ESLint has no configuration that supports the current style. Interestingly, there is an active PR to add it to ESLint: https://github.com/eslint/eslint/issues/6141 (`"multi-or-nest"` plus not enforcing unncessary parens are not used).

dcleao:
Is there any impediment for merging this @pentaho/millenniumfalcon ?

package-res/resources/web/pentaho/data/CellTuple.js
mbatchelor: I'm with Nelson on this one. As a single-line statement, it's clear that you can't add another statement inside the if bit without adding the bracket.

dcleao: Added a commit to fix both the "number of statements in a line" rule and the "multiple line statements requires braces" rule.

PR #823, opened 2016-08-18
[BACKLOG-9807] - Post-processing computation is not displayed after c…

Aliaksandr-Kastenka:
…hanging independent parameters

### DIFF
--- a/package-res/resources/web/prompting/PromptPanel.js
+++ b/package-res/resources/web/prompting/PromptPanel.js
@@ -1201,6 +1201,9 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
                   }
 
                   updateNeeded = _areParamsDifferent(dashboardParameter, paramSelectedValues, param.type);
+                  if (!updateNeeded && param.isErrorChanged) {
+                    updateNeeded = true;
+                  }
                 }
 
                 if (updateNeeded) {
--- a/package-res/resources/web/prompting/builders/DateInputBuilder.js
+++ b/package-res/resources/web/prompting/builders/DateInputBuilder.js
@@ -42,7 +42,7 @@
  *
  * @name DateInputBuilder
  * @class
- * @extends ValueBasedParameterWidgetBuilder
+ * @extends FormattedParameterWidgetBuilderBase
  */
 define(['./FormattedParameterWidgetBuilderBase', '../components/DojoDateTextBoxComponent', 'common-ui/jquery-clean'],
     function (FormattedParameterWidgetBuilderBase, DojoDateTextBoxComponent, $) {
--- a/package-res/resources/web/prompting/builders/FormattedParameterWidgetBuilderBase.js
+++ b/package-res/resources/web/prompting/builders/FormattedParameterWidgetBuilderBase.js
@@ -34,12 +34,12 @@
  *
  * @name FormattedParameterWidgetBuilderBase
  * @class
- * @extends ParameterWidgetBuilderBase
+ * @extends ValueBasedParameterWidgetBuilder
  */
-define(['common-ui/util/formatting', './ParameterWidgetBuilderBase', 'common-ui/jquery-clean'],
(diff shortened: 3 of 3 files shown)


Aliaksandr-Kastenka:
@tmorgner 
Please review.

package-res/resources/web/prompting/PromptPanel.js
```diff
@@ -1209,6 +1212,14 @@ define(['cdf/lib/Base', 'cdf/Logger', 'dojo/number', 'dojo/i18n', 'common-ui/uti
                     this.dashboard.updateComponent(component);
                   }.bind(this));
                 }
+
+                if ((param.attributes["post-processor-formula"] || param.attributes["default-value-formula"]) && param.getSelectedValuesValue().length > 0) {
```
tmorgner: This looks as if there is something else going on. The result of the post-processing is a new selected value and that value should have been present in the values array and thus be set via the "this._initializeParameterValue(null, param)" call earlier on. 

If the code above does not catch the case where the parameter-set (the elements available to be selected) remains the same, but the selection set (the values of the parameter-set that are actually selected) changes, then fix it there. I would also assume that this kind of bug would be independent of whether the parameter has a post-processing or default-value formula. 

You should be able to validate that by writing a unit-test that triggers that condition.

Aliaksandr-Kastenka:
In case when parameter has post-processing formula and display type is undefined(or textbox) "newValuesArray" is undefined
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L1176-L1179
although "param" has selected value.
This happens because "textbox" builder
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/builders/FormattedParameterWidgetBuilderBase.js#L81-L87
does not have "valuesArray" field.
"autocompletebox" does
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/builders/StaticAutocompleteBoxBuilder.js#L74-L79
@tmorgner 
Please take a look.

tmorgner:
If the text-box component has no values-array, but the code expects one, then one of two things is true:

(1) either the text-box component is wrong and should maintain a values-array 
or 
(2) the code that expects a values array on each component is wrong and should treat plain-value widgets correctly.

In either case, your narrow focus on elements that have a formula is misguided. The formulas just exposed a bug in the system, but they are not the cause of the problem in the client side. Lets fix the real problem here, which is the fact that server side changes to selected values (for whatever reason) for plain-text parameter are not reflected on an update.

Personally, I think the cleanest way of solving this would be to declare that all parameter-widgets have to have a valuesArray property at all times. It is an easy enough requirement to guarantee, has no side effects on anything else and allows us to treat all parameter-components in an uniform way. 
 
Looking at the sources, I can see multiple input widget-types that set no valueArray property when they are created:

* DateInputBuilder
* ExternalInputBuilder
* TextAreaBuilder
* TextInputBuilder

Upgrade these builders so that all of these components have a "valuesArray" property. The value should reflect the selection of the "Parameter" object (the one found in "args.param") at the time the widget is built. Make sure that - like with the other input-builders, you return a copy of the param.values array, not the original, or you might get weird side-effects later when you update the selection in your report-viewer code.

At that point you can drop the check for any of the formulas on the parameter. Your code should "just work"(tm)

Aliaksandr-Kastenka:
DateInputBuilder
ExternalInputBuilder
TextAreaBuilder
TextInputBuilder

They all extends FormattedParameterWidgetBuilderBase. Only them extends FormattedParameterWidgetBuilderBase.
So I've changed:
ParameterWidgetBuilderBase -> ValueBasedParameterWidgetBuilder
ParameterWidgetBuilderBase -> FormattedParameterWidgetBuilderBase
to:
ParameterWidgetBuilderBase -> ValueBasedParameterWidgetBuilder -> FormattedParameterWidgetBuilderBase
@tmorgner 
Please review.

tmorgner:
Looks good. And no special exception tests based on parameter-properties makes this commit a clean beauty.

PR #824, opened 2016-08-19
BACKLOG-8292 - DEV: update consuming projects if necessary

kolinus:
What was done:
changed modeler dependency


kolinus:
@rfellows modeler

rfellows:
inclusion of the gwt jar isn't required here. i'm working on modifying it

PR #826, opened 2016-08-26
[ANALYZER-3313] - Drag and drop don't work in a Measure or Calculated…

apratkin:
… Measure with name '


apratkin:
@rfellows could you please take a look?

PR #827, opened 2016-09-02
[PPP-3559] - Use of vulnerable component Angular.js 1.2.15 CVE-2015-0167

YuryBY:
@graimundo, @tmorgner, @mbatchelor, @rfellows, here is angular.js updating


lucboudreau:
@pamval Pedro, who needs to pitch in from your side on this review? This is a big upgrade.

PR #829, opened 2016-09-06
[BACKLOG-10019] Casting date and number values in google data tables.

dcleao:
Supporting Google "datetime" type as an alias to type "date"
Implemented an "as close as possible/needed" ECMA-262 v7 Date.parse in pentaho/util/date.
Changed pentaho/type/date to use the new cross-browser safe date parse function.

Related with https://github.com/pentaho/pentaho-det/pull/266.


dcleao:
Fixed duplicate member declaration I added that caused problems in stricter running environments (like PhantomJS of `ant test-js`).

dcleao:
Forced-push to make wingman run again.

PR #831, opened 2016-09-12
[BACKLOG-10029] - Transitioning from use of Agile-BI platform to supp…

rfellows:
…ort DET to the new pentaho-analyzer osgi feature.


pentaho-nbaker:
@rfellows , should we try getting Analyzer-OSGI working with the new API paths instead?

rfellows:
I assumed when i brought this up in the past, based on your reaction, that it wasn't worth the effort. now with half of these PR's merged, we need this one too.

PR #833, opened 2016-09-14
[BACKLOG-9342][BACKLOG-10514] Fixed CCC wrapper's column reordering b…

dcleao:
…ehaving unpredictably depending on column data type.

Updated CCC wrapper Line and StackedArea to support modal ordinal/quantitative X-axis visual roles
Updated Scatter and Bubble to not limiting the quantitative X and Y visual roles to the number data type, thus additionally supporting the date datatype.

Should be merged together with https://github.com/pentaho/pentaho-det/pull/305.

### DIFF
--- a/package-res/resources/web/pentaho/visual/ccc/_axes/AbstractAxis.js
+++ b/package-res/resources/web/pentaho/visual/ccc/_axes/AbstractAxis.js
@@ -21,7 +21,7 @@ define([
   "use strict";
 
   return Base.extend({
-    constructor: function(chart, axisId, gems) {
+    constructor: function(chart, axisId, mais) {
       this.chart = chart;
       this.id = axisId;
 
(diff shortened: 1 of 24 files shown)


package-res/resources/web/pentaho/visual/ccc/_axes/AbstractAxis.js
```diff
@@ -21,7 +21,7 @@ define([
   "use strict";
 
   return Base.extend({
-    constructor: function(chart, axisId, gems) {
+    constructor: function(chart, axisId, mais) {
       this.chart = chart;
```
carlosrusso: Why renaming `gem`/`gems` to  `mai`/`mais` and not the slightly longer (but more comprehensible) `mapAttrInfo`/`mapAttrInfos` ? 

I find that the shorter variant introduces a higher effort to parse and a higher brain utilization. The benefit of a short variable/method name is lost if you need to read the code around to figure out wtf is `mai`

package-res/resources/web/pentaho/visual/ccc/abstract/View.js
```diff
@@ -720,102 +836,92 @@ define([
...
-        this.options.dataCategoriesCount = null;
-      }
-    },
+        // The use of the relational translator with the following options
+        // is the most reliable way to "control" the translator code and avoid unexpected shuffling of columns
+        this.options.crosstabMode = false;
+        this.options.isMultiValued = true;
+
+        this._mappingAttrInfos.forEach(function(mai) {
+          if(mai.attr) {
```
carlosrusso: I find the usage of a short `mai` acceptable in loops like this, because the meaning of `this._mappingAttrInfos` can be understood with minor effort.

package-res/resources/web/pentaho/visual/ccc/_axes/AbstractAxis.js
dcleao: `mais` means `mappingAttributeInfos` and I have abbreviated it, in some places, as `mappingAttrInfos`. Even this abbreviation I find long to a point that it prejudices readability, so I many times used the shorted `mais`.

Like with any abbreviation, initially it is harder to read, but then it speeds up reading and comprehension.

If you look at the code from the beginning (i.e., at least, starting from the `abstract/View` file) you will find many annotations and comments that clarify the abbreviation.

dcleao:
@carlosrusso If you want, I can try to split formatting changes into a separate commit. The concept renaming is harder to separate, as that is part of the whole refactoring.

dcleao:
@carlosrusso I'd usually do a separate commit with formatting changes. I didn't do that here due to delivery pressure, so sorry it's you ending up to take more time and not me...

package-res/resources/web/pentaho/visual/ccc/_axes/AbstractAxis.js
nantunes: I agree that `mais` isn't clear. The fact it also is a portuguese word doesn't help.

dcleao:
Added a commit to improve the naming related to `MappingAttributeInfo`.

PR #835, opened 2016-09-19
[BACKLOG-10265] Promoting `dynamicAttributes` to public and moving it…

dcleao:
… to `pentaho.type.Type`.

* Adding `isVisualKey` attribute to visual roles.
* Fix definition of PointAbstract "rows" visual role as always an `isVisualKey`.
* Added new types `pentaho.type.Application` and `pentaho.type.Model`. The `pentaho.visual.base.Model` now extends from the latter.

Merge together with https://github.com/pentaho/pentaho-det/pull/328 and https://github.com/pentaho/pentaho-det-ee/pull/191.

### DIFF
--- a/package-res/resources/web/pentaho/type/model.js
+++ b/package-res/resources/web/pentaho/type/model.js
@@ -1,0 +1,98 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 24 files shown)


package-res/resources/web/pentaho/type/model.js
```diff
@@ -0,0 +1,98 @@
...
+           *
+           * @name application
+           * @memberOf pentaho.type.Model#
+           * @type {pentaho.type.Application}
+           */
+          {
+            name: "application",
+            type: "pentaho/type/application"
+          }
+        ]
```
nantunes: And the other fields referred in IModel.jsdoc? Or is it unrelated and I'm making confusion?

nantunes: Maybe I'm mixing the Model and VisualModel... But the IModel.jsdoc seems to apply to both.

PR #836, opened 2016-09-21
[BACKLOG-6859]-Row limit UI components

EgorZhuk:
@tmorgner please review


tmorgner:
I dont think wingman is running any tests here, it only tests the Java side. So there is no point in waiting, isn't there?

PR #838, opened 2016-09-22
[BACKLOG-6859]-Apply system limit as a callback for 'limit exceeded' …

EgorZhuk:
…dialog

Select 'Maximum' and trigger callback on dialog button click.

@tmorgner please review


EgorZhuk:
@tmorgner please take a look

PR #841, opened 2016-09-27
Revert "[BACKLOG-9900] - 7.0 Localization for PUC."

tmorgner:
Reverts pentaho/pentaho-platform-plugin-common-ui#840


tmorgner:
Reverting the pull request, as files have been renamed by the localisation team, but should have retained their original file name.

PR #847, opened 2016-09-29
BACKLOG-10801--DET - When we are interact with Pivot Table and then we

kolinus:
decide to resize it, the data service is called everytime we resize Spoon
window

What was done:
Added check to see if it is resize and it is DET


kolinus:
@DFieldFL @diogofscmariano  could you please review

diogofscmariano:
@pentaho/millenniumfalcon please take a look into this too :)

PR #852, opened 2016-10-06
[CLEANUP] Debuggability of Client Metadata Model and VizAPI objects

dcleao:
When the url contains "&debug=true&debugLevel=5" (or a higher debug level), a special **debug mode** is activated in the `pentaho/lang/Base.js` library.

This mode fills-in class names in constructors and method names in methods. Because of this,
instead of "[Object]" or "(anonymous function)" in a debugger watch window, something like "[pentaho_visual_barChart_model]" would show.

The debugging mode causes code to run slower.
When debugLevel is < 5, the code behaves exactly the same as it does today.

@pamval, @pentaho/millenniumfalcon please review.

ATTENTION: Only merge from Monday, 10/Oct/2016 onward.

### DIFF
--- a/package-res/resources/web/pentaho/lang/Base.js
+++ b/package-res/resources/web/pentaho/lang/Base.js
@@ -48,10 +48,12 @@
  *     Specify `extend_order` when sub-classing, e.g. `Base.extend({extend_order: ["b", "a"]});`.
  */
 define([
+  "module",
   "../util/object",
   "../util/fun",
-  "../util/text"
-], function(O, fun, text) {
+  "../util/text",
+  "../util/debugInfo"
+], function(module, O, fun, text, debugInfo) {
   "use strict";
 
   // ## Support variables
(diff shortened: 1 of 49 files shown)


package-res/resources/web/pentaho/lang/Base.js
```diff
@@ -83,6 +84,8 @@ define([
         "extend_exclude": 1
       });
 
+  var _isDebugMode = debugLevel.level > 4; // 5 is "debug"
```
carlosrusso: `debugLevel` will ultimately parse the url for  `debug=` and `debugLevel=` parameters. 
While this is the current practice, I'm twitching to the sight of this: I don't like having`pentaho.lang.Base` parse the location url, even if indirectly.

I understand that this is a practical way for a developer to globally define the debug flags, but I need some time  to overcome my prejudice.

dcleao: See my new proposal :-)

PR #854, opened 2016-10-06
[BACKLOG-8081] PIR: When using a textbox prompt, special characters a…

annapopova1:
…ppear in their encode syntax instead of their actual character


annapopova1:
I don't see broken unit tests. Wingman wrote "this pull request has passed all validations" and there are newly fixed tests.

tmorgner:
@krivera-pentaho I often get caught out by Wingman's formatting too. As the change itself looks safe and sane, lets merge.

@wingman-pentaho  can we get a different icon when a case is fixed vs newly broken? Red circle for newly broken tests, yellow for still broken, green circle for newly fixed tests? (I opened ENGOPS-2493 for it too.)

PR #859, opened 2016-10-10
[BACKLOG-9441] 'View report' button does NOT work after uncheck and t…

Aliaksandr-Kastenka:
…hen check the 'Auto-submit'

@tmorgner 
Please review.


tmorgner:
No, I'm not merging this API change. 

You cannot simply suppress the event like that. This will make the API behave completely erratic - if you don't like what happens in the event-handler, change the event handler. The event-function that is actually called is "beforeUpdateCallback" in reportviewer.js. Fix that one and check for the auto-submit status there if needed. 

You can use the prompt's operationAPI for that: "api.operation.state().autoSubmit" gives you the current status of that property.

PR #866, opened 2016-10-14
[BACKLOG-11300] - Browser freezes when putting data into disabled row…

Aliaksandr-Kastenka:
… limit textbox in

@tmorgner @EgorZhuk 
Please review.


EgorZhuk:
If it's a dojo issue and there is no other way to avoid it but hiding disabled input, please at least show a label with it's value instead.

PR #868, opened 2016-10-14
[BACKLOG-11206] - Report Viewer: 'Request failed' message when inputt…

dzmitry-hryhoryeu:
[BACKLOG-11206] - Report Viewer: 'Request failed' message when inputting letters in parameter with numeric value type.

Related pull: https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/454

@tmorgner please review.

### DIFF
--- a/package-res/resources/web/prompting/parameters/ParameterValidator.js
+++ b/package-res/resources/web/prompting/parameters/ParameterValidator.js
@@ -21,8 +21,8 @@
  * @name ParameterValidator
  * @class
  */
-define(['cdf/lib/Base'],
-    function (Base) {
+define(['cdf/lib/Base', 'common-ui/util/TextFormatter', 'common-ui/util/formatting', 'pentaho/common/Messages'],
+    function (Base, TextFormatter, FormatUtils, Messages) {
 
       var addErrorToParameter = function (paramDefn, paramName, message) {
         var errorList = paramDefn.errors[paramName];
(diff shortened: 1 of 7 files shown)


package-res/resources/web/prompting/parameters/ParameterValidator.js
```diff
@@ -105,11 +105,16 @@ define(['cdf/lib/Base'],
           // remove error, if any
           removeError(paramDefn, currentParameter.name);
 
+          var formatterType = TextFormatter.getFormatType(currentParameter.type);
+          if(formatterType == 'number' && typeof untrustedValue != formatterType) {
```
tmorgner: there can also be dates in a text field.

tmorgner:
Both number and date-fields can be validated on the client. When we send numbers or dates to the server, we rewrite  them to be in canonical format (numbers = en_US, no thousands
separators)(date = ISO-8601 format).

Therefore you have all the tools necessary on the client side to detect invalid input in text-fields as far as dates and numbers are supported. Just try to pass the inputs though the formatter's parse function to see whether you get a valid date-object. If parsing fails, if will return '' (empty string). 

Note that the parsing for numbers in those filters cannot be used to validate inputs, as there is no parsing done. The filter relies on JavaScript's automatic type-conversion for that job.   

You will still have to filter out any data-type that is not a date or number on the server though.

tmorgner:
@dzmitry-hryhoryeu Seems I lost track of this pull request after the code-freeze. 

Needs a rebase, and most definitely needs unit-tests that show that the changes detect invalid inputs reliably. Code without unit-tests cannot go into the release.

PR #869, opened 2016-10-17
[BACKLOG-11613]-Prevent changed parameters behaviour for row limit 'M…

EgorZhuk:
…aximum'

It was not a good idea to use Infinity.

@tmorgner please review


tmorgner:
Infinity was a logical choice for saying "no limit". Oh well, ...

PR #876, opened 2016-10-25
[BACKLOG-11740] - [Safari] Query limit value is not visible in the Ro…

dzmitry-hryhoryeu:
[BACKLOG-11740] - [Safari] Query limit value is not visible in the Row Limit text box.

@tmorgner please review.


tmorgner:
Oh the joy of non-conforming browsers. I feel young, its like it is 1996 again.

PR #880, opened 2016-10-27
[BACKLOG-11724] CCC wrapper - Moved CCC chart instance creation to a …

dcleao:
…separate method to allow per-instance override of the new `pvc.CategoricalChartAbstract#_onWillCreatePlotPanelScene`.

* Added validation of maximum number of plots' visual scenes.
* Removed ViewDemo.
* Forwarding any Chart#getLastRenderError() as `View#update` rejections.
* Removed clearing of the domContainer in ccc/abstract/View

Depends on https://github.com/webdetails/ccc/pull/262.

@pamval @pentaho/millenniumfalcon please review.

### DIFF
--- a/package-res/resources/web/pentaho/visual/ccc/abstract/View.js
+++ b/package-res/resources/web/pentaho/visual/ccc/abstract/View.js
@@ -28,16 +28,15 @@ define([
   "pentaho/visual/color/paletteRegistry",
   "pentaho/visual/role/level",
   "pentaho/data/TableView",
-  "./ViewDemo",
   "pentaho/i18n!view"
 ], function(View, selectionModes,
             def, pvc, cdo, pv, Axis,
             util, O, logger, visualColorUtils, visualPaletteRegistry,
-            measurementLevelFactory, DataView, ViewDemo, bundle) {
+            measurementLevelFactory, DataView, bundle) {
 
   "use strict";
 
-  /* global alert:false, cv:false */
+  /* global alert:false, cv:false, Promise:false */
 
   var ruleStrokeStyle = "#808285";  // "#D8D8D8",  // #f0f0f0
   var lineStrokeStyle = "#D1D3D4";  // "#D1D3D4"; //"#A0A0A0"; // #D8D8D8",// #f0f0f0
@@ -244,7 +243,8 @@ define([
 
       this._applyExtensions();
 
-      this._renderCore();
+      // Return any rejection promise.
+      return this._renderCore();
     },
 
     _updateSize: function() {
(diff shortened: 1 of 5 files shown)


package-res/resources/web/pentaho/visual/ccc/abstract/ViewDemo.js
```diff
@@ -126,8 +126,9 @@ define([
 
     // Temporary. Used for demo of BACKLOG-6739
     _renderCore: function() {
-      this.base();
+      var result = this.base();
```
nantunes: Why are we keeping this class?

package-res/resources/web/pentaho/visual/ccc/abstract/View.js
```diff
@@ -1382,21 +1382,58 @@ define([
...
+     * @return {number} The maximum number when defined and greater than `0`; `-1`, otherwise.
+     */
+    _getVisualElementsCountMax: function() {
 
-      var ChartClass = pvc[this._cccClass];
+      // TODO: Currently, this is DET dependent...
 
-      this._chart = new ChartClass(this.options);
+      var modelTypeApp = this.model.type.application;
+      var cellCountMax = +(modelTypeApp && modelTypeApp.optimalMaxDataSize);
```
nantunes: Ok. But then maybe we should increase the DET limit. I think 5k was chosen thinking about multiple measures, maybe 10k would be ok?

nantunes: (of course, not related with the PR)

package-res/resources/web/pentaho/visual/ccc/categoricalContinuousAbstract/View.js
```diff
@@ -67,6 +67,24 @@ define([
...
+    },
+
+    _createChart: function(ChartClass, options) {
+
+      var chart = this.base(ChartClass, options);
+
+      var visualElemsCountMax = this._getVisualElementsCountMax();
+      if(visualElemsCountMax > 0) {
+        var me = this;
+        chart.override("_onWillCreatePlotPanelScene", function(plotPanel, data, axisSeriesDatas, axisCategDatas) {
```
nantunes: How many times can this be called? (or, in another way, how many PlotPanelScene has a visualization?)

package-res/resources/web/pentaho/visual/ccc/abstract/ViewDemo.js
dcleao: Just being cautious, in case someone else uses the file.

package-res/resources/web/pentaho/visual/ccc/categoricalContinuousAbstract/View.js
dcleao: The method is called once for each plot panel that a chart has _before_ creating its scene tree. The name ain't perfect.
A chart can have more than one plot panel (e.g. column-line combo), although DET currently only exposes CCC charts with a single plot.

PR #888, opened 2016-11-03
[ANALYZER-3203] - When drilling in a Sunburst viz it adds a filter th…

stanislau-strelchanka:
[ANALYZER-3203] - When drilling in a Sunburst viz it adds a filter that was not part of the selected drill


stanislau-strelchanka:
@bennychow, @pentaho-nbaker, @rfellows, @mchen-len-son please review

PR #891, opened 2016-11-16
[CLEANUP] Debuggability of Client Metadata Model and VizAPI objects

dcleao:

### DIFF
--- a/package-res/resources/web/pentaho/lang/EventSource.js
+++ b/package-res/resources/web/pentaho/lang/EventSource.js
@@ -29,11 +29,11 @@ define([
    * @implements pentaho.lang.IEventRegistrationHandle
    * @private
    */
-  function EventRegistrationHandle(dispose) {
+  function pentaho_lang_IEventRegistrationHandle(dispose) {
     this.dispose = dispose;
   }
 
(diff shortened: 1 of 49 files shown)


package-res/resources/web/pentaho/lang/EventSource.js
```diff
@@ -119,7 +119,7 @@ define([
 
               queue[i + 1] = listenerInfo;
 
-              handles.push(new EventRegistrationHandle(removeSingleHandle.bind(this, eventType, listenerInfo)));
+              handles.push(new pentaho_lang_IEventRegistrationHandle(removeSingleHandle.bind(this, eventType, listenerInfo)));
```
dcleao: The name is chosen as giving meaning to an instance of such a class, to be read as  "an instance of a class that implements `pentaho.lang.IEventRegistrationHandle`". The class is an internal implementation detail of this module. The end dev user should not see it be used with `new`. The only public, documented concept that the user might know of is `pentaho.lang.IEventRegistrationHandle`.

Alternatively, we could give the class a local, internal name like `pentaho.lang.EventSource~EventRegistrationHandle`, which would be escaped and result in the function name identifier: `pentaho_lang_EventSource_EventRegistrationHandle`.

PR #892, opened 2016-11-17
[CLEANUP] Web Client Platform - Default Context Variables module

dcleao:
* Added “pentaho/contextVars”
* Added “pentaho/_globalContextVars”
* Added `Context.instance` singleton exposing the “default” context.
* Renamed and changed a bit the `GlobalContextVars` to `CustomContextVars`


test-js/unit/pentaho/CustomContextVars.Spec.js
```diff
@@ -0,0 +1,113 @@
...
+              "pentaho/contextVars",
+              "pentaho/CustomContextVars"
+            ], configDefaultContext, function(defaultContextVars, CustomContextVars) {
+
+              var contextVars = new CustomContextVars();
+              expect(contextVars[varName]).toBe(defaultContextVars[varName]);
+            });
+          });
+
+          it("should default to the default variable's value, when spec." + varName + " is empty", function() {
```
nantunes: We are assuming the context variables are always strings (notice at https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/892/files#diff-36fb4c51544b76dc0809d39bdc6c3175R44 that a `false` boolean value would be ignored, but a `true` would not).

But why ignore explicit empty strings? Or why can't we override a default value with `null`?

I would rather make no assumptions about the context variables and just fallback in case of `undefined`.

dcleao: Usually, I exclude the empty strings as valid values and just interpret them as null/undefined. In most contexts, there's no meaningful use for an empty string.
But usually, also, I'd document these cases as `?nonEmptyString` and it was not the case as they are documented as `?string` (https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/pentaho/_doc/spec/IContextVars.jsdoc). 
This type does not include `undefined` (would have to declare it as `string|null|undefined`).

Assuming only valid types (except `undefined`) are provided, the empty string — a member of the declared type — is converted to `null`. No attempt is made to handle values of other types (except `undefined`).

Are there any use cases for distinguishing between `null` and `undefined` (like context inheritance) ?
Do you want me to improve the code to make sure that any non-string values are converted to the default?

nantunes: For instance... suppose I want to perform something impersonating a logged out user, creating a context with `{user: null}` value would default to the current user, not to "no user".
That probably isn't the current semantics of the field, or not even a real scenario, but generically speaking seems to be a valid use case for allowing `null` values.

Omitting fields in the spec (or explicitly setting them to `undefined`, but there's no need for that) is the way I see making sense to fallback to the default values. Falling back on invalid (wrong type) or empty values seem strange to me.

nantunes:
@dcleao wingman's broken tests seem unrelated, but please check. ty

dcleao:
Rebased, tested locally and pushed force.

dcleao:
Ready for merging!

PR #894, opened 2016-11-22
[BACKLOG-11517] - PIR: An unexpected error when putting huge numbers …

dzmitry-hryhoryeu:
[BACKLOG-11517] - PIR: An unexpected error when putting huge numbers into Row Limit textbox in.


dzmitry-hryhoryeu:
@tmorgner please review.

PR #895, opened 2016-11-24
[BACKLOG-8367] Removing scroll width hack for visualizations in multi-charts "mode"

davidmsantos90:


dcleao:
Finally these parent node and scrolling hacks were removed!!

PR #898, opened 2016-11-28
[BACKLOG-9116] Rename and resize visualization icons

davidmsantos90:

Onyx:
 - icons don't expand outside the container div
 - vertical and horizontal stacked bar are now matching its visualization
 - icons size is slightly bigger in 3.0 vizs 

Crystal:
 - Icons are the same size as the 2.0 viz's icons


dcleao:
@davidmsantos90 a propos, it would be great if we'd document the standard style classes for "components" and their icons with their sizes and intented use, in `pentaho.type.Type#styleClass`. 
Including the recently defined `component-container` CSS class.

PR #904, opened 2016-12-05
[CLEANUP] Moving jsdoc files to a separate content root (like that of tests)

dcleao:
Plus:
* Fixed missing public stylesheet of generated jsdoc output
* Updated jsdoc version


dcleao:
@diogofscmariano please review.

diogofscmariano:
Good change. Why dont we just have its folder as doc-js? Shouldnt all our doc be jsdoc? :)

dcleao:
@diogofscmariano Unfortunately, the "doc-js" folder is already taken by ant/subfloor to be the output folder of generated documentation! Changing this would require changing subfloor-js...
But that was my wish, and we probably can do it like that, if appropriate, in maven projects.

diogofscmariano:
Too much time working in maven things... you are right :(

PR #905, opened 2016-12-06
[CLEANUP] Various Web Client Platform refactorings

dcleao:
## Various Web Client Platform refactorings
* Naming
  * Moved several implementation classes to `impl` subfolders.
  * Use of new naming scheme for named implementation modules, e.g. `pentaho/config/impl/instanceOfAmdLoadedService`
* Refactored the **context** types
  * Moved to `pentaho/context`
  * Making `pentaho/context` a proper JS class, and a more structured one.
  * Added most other remaining global variables that `webcontext.js` defines to `pentaho.context.spec.IVariables`.
  * Platform context info is configurable through "pentaho/context" AMD configuration
* Promoted the **config** types, from `pentaho/type/config` to `pentaho/config`
  * Global configuration service instance is available at `pentaho/service!pentaho.config.IService?single`
  * Config service no longer converts short type ids to full ids
  * Config service no longer accepts no type selector variable (used to default it to `pentaho/type/value`)
  * `IRuleSelector#type` is now required.
* Refactored the **debug** types.
  * Moved to `pentaho/debug` folder
  * Defined the `pentaho.debug.IManager` interface, which is implemented by the class `pentaho.debug.impl.Manager`,
  * Defined a system `pentaho/debug` singleton module, an instance of `pentaho.debug.impl.Manager`
  * Cannot use `pentaho/service` at this low level, as it would bring problems to AMD loader plugins, within r.js.
* `new pentaho.type.Context(contextVars)` now creates an instance of `pentaho.context.impl.CustomVariables` when receiving a plain object,
   allowing for simpler application container code.
* Various:
  * Commenting misbehaved dojo unit test.
  * Rewrote unit tests clearing the services configuration,
    because now, other unit tests depend on the default services configurations.
  * Rewrote `pentaho/service` unit tests to use the require-test helper
  * Created `pentaho/util/domWindow` and `pentaho/util/global` to be able to mock these when testing
  * Cleanup the test config scripts, after the latest conversion to maven

Merge together with:
* https://github.com/pentaho/pentaho-analyzer/pull/1359
* https://github.com/pentaho/pentaho-platform-plugin-geo/pull/134
* https://github.com/pentaho/pentaho-det/pull/579
* https://github.com/pentaho/eng-sandbox/pull/4
* https://github.com/webdetails/cdf/pull/908


package-res/resources/web/common-ui-require-js-cfg.js
```diff
@@ -63,6 +64,14 @@
   requirePaths["pentaho/_globalContextVars"] = basePath + "/pentaho/_globalContextVars";
   requirePaths["pentaho/contextVars"] = requirePaths["pentaho/_globalContextVars"];
 
+  // Singleton interface, virtual module ids
+  //requireMap["*"]["pentaho/spec/IContext"] = "pentaho/service!pentaho.spec.IContext?single";
+  requireMap["*"]["pentaho/config/IService"] = "pentaho/service!pentaho.config.IService?single";
```
nantunes: Like I told you, the module ID `pentaho/config/IService` seems strange. I like the map approach (@graimundo doesn't), but only if we could find a better nomenclature for these _default instances_.

dcleao: I also see now why it is confusing...
1. `pentaho/config/IService.instance`
2. `pentaho/config/IService.the`
3. `pentaho/config/IService.first`
4. `pentaho/config/IService.default`
5. `pentaho/config/iService` 
6. `pentaho/config/the.IService`

I'll create a poll for this on slack.

dcleao:
Rebased on master to remove from the commit list those of the meanwhile merged base PR.

doc-jsdoc/pentaho/config/spec/IRuleSelector.jsdoc
```diff
@@ -0,0 +1,74 @@
...
+ * the criteria map that determines if a rule is selected
+ * for a given type and context variables.
+ *
+ * It is the type of the [select]{@link pentaho.config.spec.IRule#select} property
+ * of a configuration rule.
+ *
+ * Besides the [type]{@link pentaho.config.spec.IRuleSelector#type} property,
+ * this interface allows filtering on any of the [context properties]{@link pentaho.spec.IContextVars}.
+ *
+ * @name pentaho.config.spec.IRuleSelector
```
dcleao: Please note "selection" to "selector" change in the rename from `IConfigurationRuleSelection` to `IRuleSelector`.

package-res/resources/web/common-ui-require-js-cfg.js
```diff
@@ -63,6 +64,14 @@
...
+
+  // Default implementations
+  requireService["pentaho/config/impl/serviceInstance"] = "pentaho.config.IService";
+  requireService["pentaho/config/impl/AmdLoadedService"] = "Class<pentaho.config.IService>";
+
```
dcleao: Of interest here, and in moved files below, is the experiment of moving auxiliary classes that implement interfaces to a sub-package named `impl`. What do you think?

package-res/resources/web/pentaho/config/util.js
```diff
@@ -0,0 +1,222 @@
...
+    "replace": mergeSpecsOperReplace,
+    "merge": mergeSpecsOperMerge,
+    "add": mergeSpecsOperAdd
+  };
+
+  /**
+   * @classDesc A singleton class that contains utilities related to configuration.
+   *
+   * @name pentaho.config.Util
+   * @class
```
dcleao: This `mergeSpecs` utility method was in the local `pentaho.type.Util` package. I ended creating another local utilities class for the config package: `pentaho.config.Util`. I pondered whether this could be a method of one of the existing utility "pentaho.util.*" but couldn't figure out a good receiver. I pondered creating a new "pentaho/util/spec" utility module, but this would put into question the current rule that "spec is a sub-package of auxiliary, JSON-like interfaces"... Then, this package is also public, and `pentaho.util.*` is almost totally private, the exception being the `MessageBundle` package. There is also no precedent of one of the utility packages in `pentaho.util` having only one public method... Big mess.

test-js/unit/pentaho/config/impl/AmdLoadedService.Spec.js
```diff
@@ -34,19 +34,25 @@ define([], function() {
         config3 = {
           rules: [
             {
-              select: {},
+              select: {
+                type: "A"
```
dcleao: `type` is now mandatory.

package-res/resources/web/pentaho/config/impl/serviceInstance.js
```diff
@@ -14,26 +14,26 @@
...
    * (see {@link pentaho.service}).
    *
    * The default system AMD registration registers the class
-   * {@link pentaho.type.config.AmdLoadedConfigurationService}.
+   * {@link pentaho.config.impl.AmdLoadedService}.
    *
-   * @name pentaho.type.configurationService
-   * @type {pentaho.type.IConfigurationService}
-   * @amd pentaho/type/configurationService
+   * @name pentaho.config.impl.serviceInstance
```
dcleao: Talvez este package devesse chamar-se `serviceDefault` ?

test-js/unit/pentaho/config/impl/AmdLoadedService.Spec.js
nantunes: An empty `type` could still mean _apply to everything_, but I'm also ok with making it mandatory.

package-res/resources/web/pentaho/config/util.js
nantunes: I think this could easily belong in `pentaho.util`. A potential receiver could be `object.js`, but I would probably prefer it to live by its own (we just need a good name). Other methods could be publicly exposed, like `cloneOwnDeep`.

Lets keep this here for now, but take a note that this should be refactored out either when similar functionality is needed elsewhere or when doing an overall revision for producing separate bundles.

package-res/resources/web/common-ui-require-js-cfg.js
nantunes: We still need to re-address this.

test-js/unit/pentaho/config/impl/AmdLoadedService.Spec.js
dcleao: Before we assumed a default of `pentaho.type.Instance`. To not have a lib specific default, I made it mandatory, but could as well have make it allow `null`. The ConfigService implementation would need to be revised, as I think it indexes rules by type. The empty string could be indexing value. 
I think that the biggest reason is that there seems to be no use case for this.
In a way, it should generally be considered a "bad practice", as we will be injecting properties into types like the JavaScript "Object".
However, if we actually start supporting types and inheritance (from `typeDecls`), "*", "" or "any" would correspond to the JavaScript's "Object" type. Specifying the latter would achieve the desired effect. (All types implicitly inherit from Object, even if not declared).

package-res/resources/web/common-ui-require-js-cfg.js
graimundo: We don't necessarily need to use the same nomenclature but on the server side that is the  guideline that is being used. We include both *api* and *impl* explicitly in the namespace (e.g. _org.pentaho.det.api.query.model.IDataSource_ / _org.pentaho.det.impl.query.model.DataSource_ ).

dcleao: I dislike the use of `api` — the regular namespaces _are the API_. The API side can contain interfaces, mandatory base classes (abstract or not), enums.
A sub-folder named `impl`, anywhere in a regular namespace, would contain auxiliary types, public and/or private, that support implementations of API types.

Ultimately, the impl subfolders could be split into a separate source root. This wouldn't map to a tree without "impl" subfolders, like would happen if "impl" could be implied... Could be used, however, to evidence separation and make it easier to distribute as separate bundles...

dcleao:
1. Rebased.
2. Added `ContextVars#basePath` to eliminate some wild CONTEXT_PATH references still around.
3. Renamed `ContextVars` and related modules to `PlatformContext` — this is to match a corresponding rename of `pentaho/type/Context` to `pentaho/type/Domain` (similar to .Net's Application Domain, or Java's Class Loader) and avoid the confusion that arises from the two contexts actually being two different things.
4. Moved `pentaho/config/util` to `pentaho/util/spec`

package-res/resources/web/common-ui-require-js-cfg.js
```diff
@@ -26,49 +29,55 @@
...
-      requireShim    = requireCfg.shim,
-      requireMap     = requireCfg.map,
+      requirePaths = requireCfg.paths,
+      requireShim = requireCfg.shim,
+      requireMap = requireCfg.map,
 
       // TODO: This fallback logic is temporary, and can be removed when the remaining
       //    parts of the system rename the "service" plugin id to "pentaho/service".
-      requireService = requireCfg.config["pentaho/service"] || (requireCfg.config["pentaho/service"] = {});
+      requireTypes = requireCfg.config["pentaho/service"] || (requireCfg.config["pentaho/service"] = {});
```
nantunes: Why the name change?

dcleao: You know :-) Just testing how the conversion to module types looks like.

dcleao:
**Summary of Changes**

1. Promoting `pentaho/type/config` to `pentaho/config`
    * Global configuration service instance available at `pentaho/service!pentaho.config.IService?single`
    * Config service no longer converts short ids to full ids
    * Config service no longer accepts no type selector variable (used to default it to pentaho/type/value)
    * Some of the renames:
        * `pentaho.type.IConfigurationService` --> `pentaho.config.IService`
        * `pentaho.type.spec.ITypeConfiguration` --> `pentaho.config.spec.IRuleSet`
        * `pentaho.type.spec.ITypeConfigurationRule` --> `pentaho.config.spec.IRule`
        * `pentaho.type.spec.ITypeConfigurationRuleSelection` --> `pentaho.config.spec.IRuleSelector`
        * `pentaho/util/spec` now contains the objects merge utility (moved code from `pentaho/type/util`)
2. Addition of `pentaho.spec.IContextVars#basePath`
    * Reads its value from the global `CONTEXT_PATH` variable.
3. `requireService = requireCfg.config["pentaho/service"]` --> `requireTypes`
4. Use of `impl` namespace for non-api stuff (e.g. `pentaho/config/impl/instanceOfAmdLoadedService`)
5. new pentaho.type.Context(contextVars) now creates an instance of CustomContextVars, when given a plain object,
   allowing for simpler application code

doc-jsdoc/pentaho/config/IService.jsdoc
```diff
@@ -15,33 +15,33 @@
...
+ * loads all registered configuration rule set modules
+ * (i.e., modules of type {@link pentaho.config.spec.IRuleSet}).
+ *
+ * The [add]{@link pentaho.config.IService#add} method
+ * is used to register a [config.spec.IRuleSet]{@link pentaho.config.spec.IRuleSet}
  * to the service.
  *
- * The [select]{@link pentaho.type.IConfigurationService#select} method
- * selects the rules of added type configurations that apply to
+ * The [select]{@link pentaho.config.IService#select} method
```
nantunes: I still prefer _extracting_ more information from the class name itself, without knowing the package. I mean `IConfigurationService`. But is just a matter of personal taste, not a blocker.

dcleao:
@nantunes Yes, some changes are still needed in other projects, like DET and Analyzer. I'll handle those as soon as we stabilize on these changes.

dcleao:
@pentaho/millenniumfalcon I rebased on master and squashed the commits — it was becoming a bit unwieldy. I apologize to those that already reviewed previous commits.
I rewrote the text on the now single commit to summarize all of the changes and also updated the PR description.
Note again that this requires some changes in other projects, which I will prepare as soon as we settle here.  Thanks!

package-res/resources/web/pentaho/util/global.js
```diff
@@ -0,0 +1,29 @@
...
+
+  // Based on ECMAScript proposal: https://github.com/ljharb/System.global
+
+  /* globals self, window, global */
+  /* eslint no-negated-condition: 0, no-new-func: 0 */
+
+  return (typeof self !== "undefined") ? self :
+         (typeof window !== "undefined") ? window :
+         (typeof global !== "undefined") ? global :
+         Function("return this")();
```
nantunes: I'm confused on what `Function(...)` does... Shouldn't it be `new Function(...)`.
I know it must be right, I simply don't understand it. 🤔

dcleao: From [MDN, on Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function):
"Invoking the Function constructor as a function (without using the new operator) has the same effect as invoking it as a constructor."

nantunes: :-D _"Hey, let's just make two of each!"
👍

package-res/resources/web/pentaho/util/spec.js
```diff
@@ -0,0 +1,222 @@
...
+    "merge": mergeOperMerge,
+    "add": mergeOperAdd
+  };
+
+  /**
+   * @classDesc A singleton class that contains utilities related to configuration.
+   *
+   * @name pentaho.util.Spec
+   * @class
+   * @amd {pentaho.util.Spec} pentaho/util/spec
```
graimundo: If they're utilities that pertain configuration shouldn't that be reflected in the namespace somehow? pentaho.util.configuration.Spec or pentaho.config.util.Spec or pentaho.config.Util or another one that refers the configuration? 
Personally I have preference for having it in the configuration namespace => pentaho.config.Util

test-js/unit/dojo/pentaho/DojoNumberSpec.js
```diff
@@ -22,10 +22,11 @@ define( ["dojo/number"], function( DojoNumber ) {
       expect( DojoNumber.parse( "1234", {} ) ).toEqual( 1234 );
     } );
 
-    it( "Dojo Number.parse(..) should accept correct locale", function() {
+    // TODO: misbehaved test; does not always work
+    xit( "Dojo Number.parse(..) should accept correct locale", function() {
```
graimundo: @davidmsantos90  was this one of the tests we were having issues with during the mavenization?

package-res/resources/web/pentaho/util/spec.js
dcleao: It's a general utility on how to merge **specification** (hence the _Spec_ name), that is used by the configuration modules but also elsewhere on pentaho/type.

graimundo: I should've take a closer look at the code instead of just looking at the class description. 
Given that info I would leave the util where it is but remove any comments that refer configuration.

doc-jsdoc/pentaho/context/spec/IVariables.jsdoc
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @name basePath
+ * @memberOf pentaho.context.spec.IVariables#
+ * @type {?nonEmptyString}
+ */
+
+/**
+ * The identifier of the client application.
+ *
+ * This is the value specified in a web context script URL, as in: `webcontext.js?context=application-id`.
```
dcleao: There is one mismatch, that I had noticed but forgot, in the meantime, which is that the Analyzer context passed to webcontext.js is "analyzer" and the application id used by Analyzer to instantiate the `pentaho/type/Context` is `pentaho/analyzer` (or something like that)...
The idea is that there is one main/default context. This context is typically an application that holds a frame. Other non-default contexts may be created as needed (they are just objects that satisfy the `pentaho.context.spec.IVariables` interface).

doc-jsdoc/pentaho/context/spec/IVariables.jsdoc
```diff
@@ -0,0 +1,119 @@
...
+ * @name user
+ * @memberOf pentaho.context.spec.IVariables#
+ * @type {?nonEmptyString}
+ */
+
+/**
+ * The home folder of the user.
+ *
+ * @name userHome
+ * @memberOf pentaho.context.spec.IVariables#
```
dcleao: I would also prefer the grouping you propose on user, reservedChars and server. The reason I did not do that was because our current configuration system does not support "path" variables on selectors.  Also, note on your URL proposal, the toString method would imply an "API" (with methods and such) and the interface is by design as simple as possible, so that it can be provided inline by webcontext.js without having to include classes, before, to do so.

doc-jsdoc/pentaho/context/spec/IVariables.jsdoc
dcleao: I've come to agree with @graimundo on keeping the two concepts separate. The main/default context will never have an application defined. Applications must create a context where they override its `application` id property. They currently already do this when creating their corresponding `pentaho/type/Context`.

doc-jsdoc/pentaho/context/spec/IVariables.jsdoc
dcleao: Ended up accepting the superior object interface. The only variable currently being used for configurations is `application`, and that name will remain unchanged. Using the `user.id` variable in the future will require support from the configuration service for selecting on "path properties".

dcleao:
@pentaho/millenniumfalcon added commit addressing @graimundo comments. Please review.

dcleao:
Rebased on master.

test-js/unit/dojo/pentaho/DojoNumberSpec.js
davidmsantos90: @graimundo only noticed this ping now, but yes, this is the same test

impl/client/src/main/doc-js/pentaho/config/spec/IRuleSelector.jsdoc
```diff
@@ -0,0 +1,74 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: Missed this license header date

impl/client/src/main/doc-js/pentaho/context/IContext.jsdoc
```diff
@@ -0,0 +1,119 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/context/IServer.jsdoc
```diff
@@ -0,0 +1,38 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/context/IUser.jsdoc
```diff
@@ -0,0 +1,42 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/context/spec/IServer.jsdoc
```diff
@@ -0,0 +1,31 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/context/spec/IUser.jsdoc
```diff
@@ -0,0 +1,39 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/debug/IManager.jsdoc
```diff
@@ -0,0 +1,102 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/debug/_namespace.jsdoc
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/doc-js/pentaho/debug/impl/_namespace.jsdoc
```diff
@@ -0,0 +1,23 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/javascript/web/pentaho/context/impl/Context.js
```diff
@@ -0,0 +1,126 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/javascript/web/pentaho/debug.js
```diff
@@ -0,0 +1,82 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/javascript/web/pentaho/debug/Levels.js
```diff
@@ -0,0 +1,112 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/javascript/web/pentaho/debug/impl/Manager.js
```diff
@@ -0,0 +1,82 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

impl/client/src/main/javascript/web/pentaho/util/spec.js
```diff
@@ -0,0 +1,222 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
```
davidmsantos90: header date update

PR #909, opened 2016-12-16
[BACKLOG-12960] - PIR: Message "Row limit value must be..." has no ba…

dzmitry-hryhoryeu:
[BACKLOG-12960] - PIR: Message "Row limit value must be..." has no background.


dzmitry-hryhoryeu:
@tmorgner please review.

PR #911, opened 2016-12-20
[BACKLOG-12872] Fix generic measures in tooltip, Column-Line combo an…

dcleao:
…d Pie chart percentage value format.


dcleao:

PR #913, opened 2016-12-22
[PRD-5793] - PRPT reports with auto-submit=false do not display on a …

dzmitry-hryhoryeu:
[PRD-5793] - PRPT reports with auto-submit=false do not display on a Dashboard.


dzmitry-hryhoryeu:
@tmorgner please review.

dzmitry-hryhoryeu:
@tmorgner please review.

PR #918, opened 2017-01-07
[BACKLOG-12457] Added the notion of view state to pentaho.visual.base…

carlosrusso:
….View

@pentaho/millenniumfalcon Please review

### DIFF
--- a/package-res/resources/web/pentaho/visual/base/View.js
+++ b/package-res/resources/web/pentaho/visual/base/View.js
@@ -80,8 +80,9 @@ define([
      * @description Initializes a `View` instance.
      * @param {!DOMElement} domContainer - The container element.
      * @param {!pentaho.visual.base.Model} model - The visualization model.
+     * @param {state} model - The initial state of the view.
      */
-    constructor: function(domContainer, model) {
+    constructor: function(domContainer, model, state) {
       if(!domContainer) throw error.argRequired("domContainer");
       if(!model) throw error.argRequired("model");
 
@@ -102,6 +103,15 @@ define([
        * @private
        */
       this.__model = model;
+
+      /**
+       * The state of the view.
+       *
+       * @type {object}
+       * @readOnly
+       * @private
+       */
+      this.__state = state;
 
       /**
        * The promise for the completion of the current update operation, if any; or `null`.
@@ -972,6 +982,26 @@ define([
      * @see pentaho.visual.base.View#_updateAll
      */
     // endregion
+  })
+  .implement(/** @lends pentaho.visual.base.View# */{
+
+    /**
+     * Gets the state of the view as a plain object.
+     *
+     * @return {object}
+     */
+    toJSON: function() {
+      return this.__state || null;
+    },
+
+    /**
+     * Gets the state of the view.
+     *
+     * @return {object}
+     */
+    getState: function() {
+      return this.toJSON();
+    }
   });
 
   /**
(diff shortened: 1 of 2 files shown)


test-js/unit/pentaho/visual/base/View.Spec.js
```diff
@@ -1164,6 +1171,53 @@ define([
...
+
+      it("should return the state of the view", function() {
+
+        var state = {};
+        var view = new DerivedView(document.createElement("div"), model, state);
+
+        expect(view.getState()).toBe(state);
+      });
+
+    }); // #toJSON
```
davidmsantos90: this should be // #getState

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -80,8 +80,9 @@ define([
      * @description Initializes a `View` instance.
      * @param {!DOMElement} domContainer - The container element.
      * @param {!pentaho.visual.base.Model} model - The visualization model.
+     * @param {state} model - The initial state of the view.
```
davidmsantos90: you have the name of the parameter inside the type slot.

@param {type} state - ...

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -104,6 +105,15 @@ define([
       this.__model = model;
 
       /**
+       * The state of the view.
+       *
+       * @type {object}
+       * @readOnly
+       * @private
+       */
+      this.__state = state;
```
dcleao: I don't think state should be stored by the base class — it is "abstract" in this respect.
It is up to specific view sub-classes to "parse" it and store it somehow.

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -80,8 +80,9 @@ define([
      * @description Initializes a `View` instance.
      * @param {!DOMElement} domContainer - The container element.
      * @param {!pentaho.visual.base.Model} model - The visualization model.
+     * @param {object} state - The initial state of the view.
```
dcleao: `@param {Object} [state] - The initial state.`

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -972,6 +982,26 @@ define([
      * @see pentaho.visual.base.View#_updateAll
      */
     // endregion
+  })
+  .implement(/** @lends pentaho.visual.base.View# */{
+
+    /**
```
dcleao:
```js
/**
 * Gets the state of the view as a plain object, if any.
 *
 * This method simply calls the [getState]{@link pentaho.visual.base.View#getState} method
 * and provides support for `JSON.stringify`.
 *
 * @return {Object} The state of the view, or `null`, if none.
 *
 * @see pentaho.visual.base.View#getState
 */
```

Also, please, call `getState` from this method, and not the other way round.

package-res/resources/web/pentaho/visual/base/View.js
```diff
@@ -972,6 +982,26 @@ define([
...
+    toJSON: function() {
+      return this.__state || null;
+    },
+
+    /**
+     * Gets the state of the view.
+     *
+     * @return {object}
+     */
+    getState: function() {
```
dcleao:
```js
/**
 * Gets the state of the view as a plain object, if any.
 *
 * The object returned by this method can later be given to the view's constructor, in argument `state`.
 *
 * The returned object must be JSON serializable (must have no cycles, and can only contain numbers, booleans, strings, plain objects and arrays).
 *
 * This implementation simply returns `null`. Override this method to expose a view's specific state. When overriding, take care to respect a state object that may be returned by the base class implementation.
 *
 * @return {Object} The state of the view, or `null`, if none. 
 * @see pentaho.visual.base.View#toJSON
 */
```

Please, just return `null`.

PR #919, opened 2017-01-09
[BACKLOG-13170]-Pass widget id to parameterChanged callback in order …

EgorZhuk:
…to have access to the widget

Currently we can't access a parameter widget in a parameterChanged callback.
Need it for https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/512


EgorZhuk:
@tmorgner updated the pull request

PR #921, opened 2017-01-12
[BISERVER-13280]-Support blur and click at the same time

EgorZhuk:
When a user types a value into a mandatory textfield and clicks the 'View report' button, 
the blur event prevents the click event on the button from being fired, so the user is forced to click it again.
A mousedown event is not prevented though, so we can set a timeout for the click logic execution.
If no click event happens the timeout will fire a submit, if it still happens ( for examle the user just clicked the button and no mandatory prompts were involved ) then the timeout is cleared and we are safe from multiple executions.


EgorZhuk:
The Jenkins build is OK, so failed junits are probably caused by a cosmic rays.

PR #923, opened 2017-01-17
[BACKLOG-13941] Fixed the masks of the values of the slices in pie ch…

carlosrusso:
…arts.

I followed an holistic approach while solving this bug, and as such I'm not entirely sure this is the best solution.


lucboudreau:
Are these failures normal? Does it often fail randomly like this?

graimundo:
@lucboudreau unfortunately yes :( I've identified the tests that are failing randomly during the mavenization of common-ui and opened a jira for it: http://jira.pentaho.com/browse/BACKLOG-13450

PR #924, opened 2017-01-17
[BACKLOG-13170]-Sync parameter definition with latest user selection

EgorZhuk:
Hey @tmorgner 
Your idea with a transient parameter didn't worked well, datepickers chnges caused an infinite loop, lists were not updated unless we requested it by a special flag, also we needed to hide the transient value...
Nevertheless the place for the fix was defined correctly.
What we are doing here - detect if the dashboard and parameter definition are not in sync, update the latest, request a parameter component update, thats it.
Checked for various controls and datatypes, see the params.prpt in the Jira ticket.


tmorgner:
Sweet work here!

PR #929, opened 2017-01-23
[PRD-5793] - PRPT reports with auto-submit=false do not display on a …

EgorZhuk:
…Dashboard.

Was fixed in 6.1 but we shouldn't forget that async execution which was introduced to 7.0 is switchable, so we also need it in master in case user disabled async mode.


EgorZhuk:
@mbatchelor fixed

EgorZhuk:
[PRD-5793](https://trello.com/c/eCeAxxXN/3-prd-5793)

EgorZhuk:
@mbatchelor @pamval @tmorgner this one is also OK now

PR #932, opened 2017-01-26
[BACKLOG-14238]-Move synchronisation of a parameter definition and a …

EgorZhuk:
…dashboard to the reporting plugin

Fixes regression in PIR by moving logic to the reporting plugin


tmorgner:
@EgorZhuk can you have a look what makes the PhantomJS tests fail so misserably? Is it wingman or a genuine issue here?

EgorZhuk:
@tmorgner currently working on it, the latest master also fails locally, so need some time to understand what is going on

EgorZhuk:
@tmorgner fixed the tests, lets wait for a wingman

PR #935, opened 2017-01-30
[BACKLOG-13280]-Ensure only one click is needed to submit a report

EgorZhuk:
Still needed a second click when the 'click' event was fired and the timeout was removed - for some reason the behaviour defined in a parent control was not applied. So, we are unbinding all the handlers and apply our simple logic:
on 'mousedown' 
- call expressionStart
- set timeout for the expresiion call
on 'click' 
- if there is no timeout set, call expression
on timeout 
- call expression and delete a timeout from 'this'

Tested in Chrome, FF and IE


tmorgner:
Oh the weird tricks we play to make browsers run fine.

PR #938, opened 2017-02-02
[PRD-5793]-Don't fail if there is no prompt UI on the page

EgorZhuk:
The previous aproach didn't worked in 7+ with async mode = false


tmorgner:
There are days where I wished we had a full-scale deep test suite that actually covers all the weird little bits, side effects and other creepy things that go on, go in, and come out of the code. Oh well, wishing is futile I guess.

Merged.

PR #940, opened 2017-02-14
[BISERVER-13522] Migrated or Imported Interactive reports do not run …

mchen-len-son:
…due to missing XMI extension

- added fallback function to cover the case when report is referencing "old" datasource"s domain ID

### DIFF
--- a/impl/client/src/main/javascript/web/dataapi/controller.js
+++ b/impl/client/src/main/javascript/web/dataapi/controller.js
@@ -70,7 +70,38 @@ pentaho.pda.app.prototype.discoverSources = function(callback, options) {
 	return sources;
 }
 
+/**
+ * This function is calling a datasource search once again,
+ * falling back to a new ID (including ".xmi" extension).
+ * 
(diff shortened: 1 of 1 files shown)


mchen-len-son:
This is a hotfix for regression found (and fixed, but as we realized, incompletely).

Scenario: 
 - datasource and PIR report are exported from BA server of an older version (we used 5.4).
 - datasource is imported to newer version, and is saved containing "xmi" extension in domain ID.
 - report is uploaded to the repository, but is still referencing to an older domain ID.

I added a fallback to a "new" domain ID, when old one is not found, like it was done in https://github.com/pentaho/pentaho-metadata/pull/139.

As I can see (https://github.com/search?p=2&q=org%3Apentaho+app.getSources&type=Code), this function is called from many places, so I decided to put the changes here.
The changes are not breaking, and every consumer of getSources function should get the same results as it did before, except the described scenario.

@pamval @mbatchelor can you please take a look?

PR #941, opened 2017-02-14
[SP-3297][BISERVER-13522] Migrated or Imported Interactive reports do…

mchen-len-son:
… not run due to missing XMI extension

- added fallback function to cover the case when report is referencing "old" datasource"s domain ID


mchen-len-son:
This is a backport of #940.

PR #950, opened 2017-02-16
[BACKLOG-14522] Web client configuration bundle

nantunes:
@pentaho/millenniumfalcon please take a look. It's the configuration bundle that will get exploded to `system/karaf/config` by https://github.com/pentaho/pentaho-karaf-assembly/pull/277.

~~The deployment of the artifact with classifier (and no primary artifact) by maven-bundle-plugin:2.5.3:install (default-install) generates an `[ERROR] No artifact found, try "mvn install bundle:install"` but it works (should be a WARNING or just INFO like maven-install-plugin:2.5.2:install (default-install) does).~~ (fixed by 2e581e2)

Check the comments for other notes.


nantunes:
One thing here is to define the default content of the `config.js` file.

Remember that, as it is, it will be the same in both PDI and server. If not instead we'll need to create separate configuration bundles (probably in DET and Analyzer?).

What do you think?

PR #955, opened 2017-02-22
[BACKLOG-13668] VizAPI View class restructuring

dcleao:
* Updated VizAPI core classes
* Updated CCC wrapper visualizations
* Updated Calculator sample visualization
* Updated sample Sandbox container
* Moved "extension" from Model to View
* Fixed Base.js overriding functions set on setter properties
* Viz model and view Id changes
* Removed Model#isInteractive
* Renamed the app-id of Analyzer
* Activated filtering for the common-ui-require-js-cfg.js file,
  to support the custom GEO AMD map configurations

Merge together with:
1. https://github.com/pentaho/pentaho-analyzer/pull/1399
2. https://github.com/pentaho/pentaho-det/pull/623
3. https://github.com/pentaho/pentaho-platform-plugin-geo/pull/154
4. https://github.com/webdetails/cdf/pull/913
5. https://github.com/webdetails/cde/pull/768
6. https://github.com/pentaho/eng-sandbox/pull/5


impl/client/src/main/javascript/web/pentaho/visual/base/view.js
```diff
@@ -0,0 +1,1322 @@
...
+       * Called when the view is constructed.
+       *
+       * Override this method, from a configuration mixin class,
+       * to perform any initialization.
+       *
+       * @param {pentaho.visual.base.spec.IViewEx} viewSpec - The view specification provided at construction time,
+       * if any.
+       * @protected
+       */
+      _init: function(viewSpec) {
```
nantunes: Should the _"overridable"_ methods be marked as private (starting with `_`)?

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
```diff
@@ -0,0 +1,1322 @@
...
+       * Called when the view properties have changed.
+       *
+       * The default implementation marks the view as dirty.
+       * More specifically, it marks the _property groups_ of the properties affected by the given changeset as dirty.
+       *
+       * The recognized property groups are those of
+       * [View.PropertyGroups]{@link pentaho.visual.base.View.PropertyGroups}.
+       *
+       * Implementations can override this method to change the default behavior for some or all of the
+       * model's properties.
```
nantunes: Not only model properties, but also custom state and view properties.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
```diff
@@ -0,0 +1,1322 @@
...
+       * @rejects {pentaho.lang.ArgumentRequiredError} When `viewSpec` is not specified.
+       * @rejects {pentaho.lang.ArgumentRequiredError} When `viewSpec` has no annotated type, inline, and
+       * the `model` property is unspecified.
+       * @rejects {pentaho.lang.ArgumentRequiredError} When `viewSpec` has no annotated type, inline, and
+       * the `model` property is a specification which also does not have its type annotated inline.
+       * @rejects {Error} When there isn't a registered default view class for the type of `model`.
+       * @rejects {Error} When the registered default view class does not exist or otherwise fails to load
+       * by {@link pentaho.type.Context#getAsync}.
+       */
+      createAsync: function(viewSpec) {
```
nantunes: Can't we remove all the _"Async"_ suffixes and just assume an asynchronous interface? I find it cleaner to have things like `View.create(spec)`, `View.getClassFromModel(model)`, ...

impl/client/src/main/javascript/web/pentaho/lang/Base.js
```diff
@@ -920,7 +920,14 @@ define([
     /* jshint validthis:true*/
 
     if(fun.is(value)) {
-      this[name] = methodOverride(value, this[name], rootProto, name);
+      // Only override if it is a normal property.
+      // When there is not baseDesc, methodOverride may need to provide a default base method.
+      var baseDesc = O.getPropertyDescriptor(this, name);
+      if(baseDesc && (baseDesc.get || baseDesc.set)) {
```
dcleao: No, this only applies when the _source_ property is *not* a getter/setter — which is how you'd override a getter/setter property —, but the target property is.

impl/client/src/main/javascript/web/pentaho/type/instance.js
```diff
@@ -253,7 +253,7 @@ define([
       _extend: function(name, instSpec, classSpec, keyArgs) {
         var typeSpec;
         if(name == null && (typeSpec = (instSpec && instSpec.type))) {
-          name = typeSpec.id || typeSpec.sourceId || null;
+          name = typeSpec.sourceId || typeSpec.id || null;
```
dcleao: The reason is that typically, the `sourceId` is a more faithful source for the name of a class. E.g. `pentaho/visual/base` vs `pentaho/visual/base/model`; the former would generate a class name of `pentaho.visual.Base` and the latter a class name of `pentaho.visual.base.Model`.
I'll update the comment.

impl/client/src/main/javascript/web/pentaho/visual/ccc/barAbstract/model.js
```diff
@@ -18,24 +18,25 @@ define([
   "../categoricalContinuousAbstract/model",
   "pentaho/i18n!../abstract/i18n/model",
   "../abstract/mixins/settingsMultiChartType"
-], function(module, categoricalContinuousAbstractModelFactory, bundle, settingsMultiChartType) {
+], function(module, baseModelFactory, bundle, settingsMultiChartType) {
```
dcleao: That's a good point. How do you propose things could be structured?
`pentaho/visual/stock/bar` and `pentaho/visual/stock/bar/ccc/view` ?

impl/client/src/main/doc-js/pentaho/visual/base/spec/IView.jsdoc
```diff
@@ -0,0 +1,75 @@
...
+ * @see pentaho.visual.base.View#height
+ */
+
+/**
+ * The usage intention of the view, namely if interaction is enabled or not.
+ *
+ * For example, a container application can specify a value of `false`
+ * when rendering a view for printing purposes.
+ *
+ * @name isInteractive
```
dcleao: I would then propose to simply remove the property for the beta and think of a better alternative for later. We can test if currently setting it to `true`, when printing, comes ith any actual harm.

impl/client/src/main/doc-js/pentaho/visual/base/spec/IViewEx.jsdoc
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The `IViewEx` interface adds to the [IView]{@link pentaho.visual.base.spec.IView} interface
```
dcleao: Yes, that's the usual meaning in a programming context,, afaik.

impl/client/src/main/doc-js/pentaho/visual/base/spec/IViewEx.jsdoc
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @name pentaho.visual.base.spec.IViewEx
+ * @interface
+ * @extends pentaho.visual.base.spec.IView
+ */
+
+/**
+ * The container DOM element where the view is rendered.
+ *
+ * @name domContainer
```
dcleao: The base interface `spec.IView` is not what you specify in the constructor but what you obtain from serialization. The `spec.IViewEx` adds to this interface additional arguments you can specify along with the serialization spec to initialize the view.

impl/client/src/main/doc-js/pentaho/visual/base/spec/IViewType.jsdoc
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+/**
+ * Extension properties that a view knows how to handle directly.
+ *
+ * This property is expected to be specified through configuration.
+ *
+ * Custom view classes should document the supported options.
+ *
+ * @name extension
```
dcleao: The `spec.IViewEx` contains instance data, while `extension` is type level metadata. Also the `Ex` in `spec.IViewEx` does not have this meaning — it's just an extended version of the base interface for use in constructors.

impl/client/src/main/javascript/web/pentaho/visual/base/model.js
```diff
@@ -299,39 +290,20 @@ define([
...
             isRequired: true
           },
-          {
-            name: "selectionFilter",
-            type: "pentaho/type/filter/abstract",
-            value: {_: "pentaho/type/filter/or"},
-            isRequired: true
-          },
+
+          // TODO: remove from here
```
dcleao: I settled with @graimundo to do this upon the interaction story.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
dcleao: The actual, more perfect, convention, that we have failed to apply consistently, is to use `__` for private and `_` for protected. This is one of the few classes where we have done this consistently. Some `pentaho/type` classes also do this consistently. I'd love a story to fix this.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
dcleao: When methods are either sync or async, that's preferable, I guess. However, because we have `pentaho.type.Type#create` and `pentaho.type.Type#createAsync`, `pentaho.type.Context#get` and `pentaho.type.Context#getAsync`, and `pentaho.type.Context#getAllAsync`, side-by-side, it would be inconsistent to have an asynchronous`pentaho.visual.base.View.create` method that just slightly changes the semantics of  `pentaho.type.Type#createAsync`.

impl/client/src/main/javascript/web/pentaho/visual/ccc/barAbstract/model.js
nantunes: Maybe keep `ccc` as a top umbrella folder... So they can possible be distributed in a separate bundle.

I think this can be a great move! It would better explain the _why_ of the model / view separation.

I even think our doc examples should start with an example only implementing a view and reusing an existing model, and a more advanced example where the developer needs to create its own model.

(in fact, as we don't have a viz registry but a model registry, both need to have their own model so the viz appears? but the first can simply extend one?)

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
nantunes: Oh, wasn't even aware of that. Ok.

impl/client/src/main/doc-js/pentaho/visual/base/spec/IViewEx.jsdoc
nantunes: Just clarifying, because of my doubts bellow. 😉

impl/client/src/main/javascript/web/pentaho/visual/ccc/barAbstract/model.js
dcleao: models: `pentaho/visual/common/bar` and views: `pentaho/visual/ccc/bar` ?
Until we have a more powerful service registry, though, we'd need to still configure the `defaultView` of  `pentaho/visual/common/bar` to `pentaho/visual/ccc/common/bar`.

dcleao: Settled on:
* `pentaho/visual/models/<bar>`
* `pentaho/visual/ccc/views/<bar>`

assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
```diff
@@ -270,28 +272,28 @@
...
-    "pentaho/visual/ccc/metricDotAbstract",
-    "pentaho/visual/ccc/areaStacked",
-    "pentaho/visual/ccc/pie",
-    "pentaho/visual/ccc/heatGrid",
-    "pentaho/visual/ccc/sunburst",
-    "pentaho/visual/ccc/donut",
-    "pentaho/visual/ccc/scatter",
-    "pentaho/visual/ccc/bubble"
-  ].forEach(registerVizPackage);
+    "pentaho/visual/models/abstract",
```
davidmsantos90: This comment should also be changed. Following the change in resources/config.js:
`Pentaho stock visualizations' models`

PR #961, opened 2017-03-08
[BACKLOG-14836] Added support for type alias and property alias.

carlosrusso:
Marked pentaho.type.element as abstract. Updated documentation.
Added unit tests around instantiation and serialisation of filters using aliases.
Replaced the `pentaho.type.Instance#toSpec`’s option, `includeType`, by two equivalent options, `declaredType` and `forceType`.
Implemented special number, string and boolean serialization under an abstract declaredType.

### DIFF
--- a/impl/client/src/main/doc-js/pentaho/type/spec/IPropertyTypeProto.jsdoc
+++ b/impl/client/src/main/doc-js/pentaho/type/spec/IPropertyTypeProto.jsdoc
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ * Copyright 2010 - 2017 Pentaho Corporation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,6 +63,26 @@
  * @type {!nonEmptyString | any}
  *
  * @see pentaho.type.Property.Type#name
+ */
+
+/**
+ * The alias for the name of the _property type_.
+ *
+ * The alias of a _property type_ is an alternative name for serialization purposes.
+ *
+ * ### Set
+ *
+ * This attribute can only be set when defining a new _property type_,
+ * and cannot be changed afterwards.
+ *
+ * When set to a non-{@link Nully} and non-{@link String} value,
+ * the value is first replaced by the result of calling its `toString` method.
+ *
+ * @name nameAlias
+ * @memberOf pentaho.type.spec.IPropertyTypeProto#
+ * @type {!nonEmptyString | any}
+ *
+ * @see pentaho.type.Property.Type#nameAlias
  */
 
 /**
--- a/impl/client/src/main/javascript/web/pentaho/type/Context.js
+++ b/impl/client/src/main/javascript/web/pentaho/type/Context.js
@@ -49,16 +49,7 @@ define([
   // Default `base` type in a type specification.
   var _defaultBaseTypeMid = "complex";
 
-  // Standard types which can be assumed to already be loaded.
-  var _standardTypeMids = {};
-
-  Object.keys(standard).forEach(function(name) {
-    if(name !== "facets" && name !== "filter") _standardTypeMids[_baseMid + name] = 1;
-  });
-
-  Object.keys(standard.facets).forEach(function(name) {
-    _standardTypeMids[_baseFacetsMid + name] = 1;
-  });
+  var O_hasOwn = Object.prototype.hasOwnProperty;
 
   var Context = Base.extend(module.id, /** @lends pentaho.type.Context# */{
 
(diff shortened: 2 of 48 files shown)


carlosrusso:

impl/client/src/main/javascript/web/pentaho/type/property.js
```diff
@@ -280,6 +281,47 @@ define([
...
+        },
+
+        set nameAlias(value) {
+          value = nonEmptyString(value);
+
+          if(!value) throw error.argRequired("nameAlias");
+
+          // Hierarchy consistency
+          var baseValue = this._nameAlias;
+          if(baseValue && value !== baseValue)
```
nantunes: But it can change it if undefined in the base? Can it can raise any inconsistency?

carlosrusso: Indeed we seem to allow it. Note that it follows the same logic used for #name.

carlosrusso: Well, the value is not inherited from the base as per https://github.com/carlosrusso/pentaho-platform-plugin-common-ui/blob/feature/BACKLOG-14836/impl/client/src/main/javascript/web/pentaho/type/_type.js#L108-L113

dcleao: The name of a property is settled on the _root_ property type `R` and cannot change in sub property types, `A` and `B`.
Name aliases, however, can be set on the root or only on subtrees, and still cannot change in that subtree once set. This means that a serialization of a type `A` is not necessarily understood by type `B`, even though a property is defined in `R`. Given a spec of `B`, one must first name aliases into actual names of `R` and only then can the spec be correctly read by `A`. I think this is an acceptable limitation, but if you guys think this brings undesirable concerns at the cost of a neglectable increase of functionality, then we can force the alias to only be settable at the root property.

impl/client/src/main/javascript/web/pentaho/type/filter/_core/tree.js
```diff
@@ -193,6 +193,7 @@ define([
         props: [
           {
             name: "operands",
+            nameAlias: "args",
```
dcleao: I still find it confusing to have a name and a nameAlias that don't share a radical word.

impl/client/src/main/javascript/web/pentaho/type/filter/_core/not.js
```diff
@@ -113,12 +113,14 @@ define([
...
       type: /** @lends pentaho.type.filter.Not.Type# */{
         id: "pentaho/type/filter/not",
+        alias: "not",
 
         styleClass: "pentaho-type-filter-not",
 
         props: [
           {
             name: "operand",
+            nameAlias: "arg",
```
dcleao: I still find it confusing to have a name and a nameAlias that don't share a radical word.

impl/client/src/main/doc-js/pentaho/type/spec/IPropertyTypeProto.jsdoc
```diff
@@ -66,6 +66,26 @@
  */
 
 /**
+ * The alias of the _property type_.
+ *
+ * The alias of a _property type_ is an alternative identifier for serialization purposes.
+ *
+ * ### Set
+ *
+ * This attribute must be set when defining a new _property type_,
```
carlosrusso: :+1:

impl/client/src/main/doc-js/pentaho/type/spec/IPropertyTypeProto.jsdoc
```diff
@@ -66,6 +66,26 @@
  */
 
 /**
+ * The alias of the _property type_.
+ *
+ * The alias of a _property type_ is an alternative identifier for serialization purposes.
```
carlosrusso: :+1:

impl/client/src/main/doc-js/pentaho/type/spec/IPropertyTypeProto.jsdoc
```diff
@@ -66,6 +66,26 @@
  */
 
 /**
+ * The alias of the _property type_.
```
carlosrusso: :+1:

impl/client/src/main/javascript/web/pentaho/type/Context.js
```diff
@@ -953,6 +948,8 @@ define([
           if(config) type.constructor.implement(config);
 
           this._byTypeId[id] = InstCtor;
+          var alias = type.alias;
+          if(alias && !O.getOwn(this._byTypeId, alias)) this._byTypeId[alias] = InstCtor;
```
carlosrusso: My original intention would actually be better expressed if I had used `O_hasOwn` instead of `O.getOwn`. We should not need to check the value at all, just the existence of a key.

impl/client/src/main/javascript/web/pentaho/type/filter/_core/tree.js
carlosrusso: "o"

impl/client/src/main/javascript/web/pentaho/type/filter/_core/not.js
carlosrusso: "o"

dcleao:
@nantunes please note that the alias change to `o` requires a server-side counterpart change in DET

PR #962, opened 2017-03-09
[BACKLOG-14324] VizAPI actions

dcleao:
@pentaho/millenniumfalcon, please review.

Depends on https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/962.


dcleao:
@pentaho/millenniumfalcon Just a notice that I added a new commit with the changes to the event classes.

dcleao:
@pentaho/millenniumfalcon Just a notice that I added a new commit with the VizAPI action classes!

dcleao:
@pentaho/millenniumfalcon These commits finish up the raw coding work. Now need a week of testing and still some docs.

assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
```diff
@@ -38,6 +38,7 @@
   // TODO: This fallback logic is temporary, and can be removed when the remaining
   //    parts of the system rename the "service" plugin id to "pentaho/service".
   var requireTypes = requireCfg.config["pentaho/service"] || (requireCfg.config["pentaho/service"] = {});
+  var requireTypeInfo = requireCfg.config["pentaho/typeInfo"] || (requireCfg.config["pentaho/typeInfo"] = {});
```
nantunes: Why is this needed in this story?

impl/client/src/main/javascript/web/pentaho/type/action/base.js
```diff
@@ -51,10 +51,43 @@ define([
...
+         * When the action is [asynchronous]{@link pentaho.type.action.Base.Type#isSync},
+         * this method _may_ return a promise. If the promise is rejected,
+         * the action is rejected with the rejection reason.
+         * However, if the promise is fulfilled, its value is always *ignored*.
+         *
+         * @param {pentaho.type.action.Base} action - The action.
+         *
+         * @return {?Promise} - A promise for the completion of default action of an asynchronous action, or `null`.
+         */
+        defaultAction: function(action) {
```
nantunes: The default action is so well separated I wonder why it exists.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
```diff
@@ -729,6 +735,95 @@ define([
...
+      },
+      // endregion
+
+      // region Actions
+      __emitActionPhase: function(action, phase, isFinal) {
+        var eventType = action.type.id;
+
+        // TODO: emitGenericAsync when action is async.
+
+        this._emitGeneric(action, eventType, phase, isFinal ? null : _emitActionKeyArgs);
```
nantunes: Where is `_emitGeneric` defined?

nantunes: We might need to emit an event payload parallel to the action (like the original DOM event that triggered the action). Where do you see it fitting better?

impl/client/src/main/javascript/web/pentaho/type/action/base.js
dcleao: Actually, I think this should be moved to beside the instance side protected methods.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
dcleao: It's `EventSource#_emitGeneric`.

dcleao: In principle, the action replaces the old event, for all uses. For that one, not sure. If, clearly, a DOM event triggered the action, then an action property would make sense to hold it. This would support async actions as well (once the DOM event is always accessible at window.event anyway).

assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
dcleao: This two lines, specifically:
```js
requireTypeInfo["pentaho/visual/action/select"] = {alias: "select"};
requireTypeInfo["pentaho/visual/action/execute"] = {alias: "execute"};
```
 allow attaching to action events using short ids. Check how `EventSource` resolves given eventType names by passing them through the `pentaho/typeInfo` service.

The other `pentaho/type` declarations in this file, I just added them for completeness.

impl/client/src/main/javascript/web/pentaho/type/action/base.js
dcleao: And, it exists for implementing the actions' default behavior. Check SelectAction's implementation.

impl/client/src/main/javascript/web/pentaho/visual/base/view.js
nantunes: Sorry, the diff was hidden when I did the search.

PR #974, opened 2017-03-19
[BACKLOG-14324] Implemented `pentaho.type.Filter#toDnf` and several supporting features and optimizations.

dcleao:
* `pentaho.type.filter.Abstract#contentKey` — allows comparing two filters for equality using a string.
* `pentaho.type.filter.Abstract#toDnf` — Converts any filter to DNF form. On certain filters, the algorithmic complexity reveals itself and blows up memory.
* `pentaho.type.filter.Abstract#andNot` — Optimized method for "subtracting" two filters that allows escaping the explosion of `(A and not B).toDnf() and provides an acceptable selection user experience.
* `pentaho.visual.action.SelectionModes#toggle` no longer needs to use to the model's data to determine if all of the input selection is already included in the current selection — it now does so extensionally (and so whether data is in crosstab or not is no longer relevant for filters).
* `pentaho.type.filter.True` and `pentaho.type.filter.False` — Along with singleton instances.
* `pentaho.type.filter` classes are now immutable, so no preventive clones need to be performed.
* `pentaho.type.filter` classes are now, each, a one type aggregate, so no preventive clones need to be performed to prevent forming inverse, bubbling references between filter objects — the cause of severe memory leaks in the `toDnf` algorithm.
* Feature — Renamed `pentaho.type.Property.Type#isReadOnly` to `isEnabled` to free the old name for another (the following) feature.
* Feature — Immutable Containers: `Complex` and `List`. Added `pentaho.type.Property.Type#isReadOnly` and `pentaho.type.List#isReadOnly`.
* Feature — Aggregates. The new attributes `pentaho.type.Property.Type#isBoundary` and `pentaho.type.List#isBoundary` allow controlling the extent of "aggregates", which are delimited by boundary properties. Change events bubble, and validation occurs, only within objects of the same aggregate. Also, objects of an aggregate prevent each other from being garbage collected (due to inverse references, that support bubbling).

Merge https://github.com/pentaho/pentaho-analyzer/pull/1432 afterwards.

### DIFF
--- a/impl/client/src/test/javascript/pentaho/type/filter/abstract.spec.js
+++ b/impl/client/src/test/javascript/pentaho/type/filter/abstract.spec.js
@@ -193,7 +193,7 @@ define([
         });
       });
 
-      it("should convert an empty `and` to false", function() {
+      it("should convert an empty `and` to true", function() {
 
         expectDnf({
           _: "and"
--- a/impl/client/src/test/javascript/pentaho/type/filter/abstract.spec.js
+++ b/impl/client/src/test/javascript/pentaho/type/filter/abstract.spec.js
@@ -166,5 +166,250 @@ define([
       });
     }); // #visit
 
+    fdescribe("#toDnf()", function() {
+
+      function createFilter(spec) {
+        return AbstractFilter.type.create(spec);
+      }
+
+      function specToDnf(spec) {
+        return createFilter(spec).toDnf();
+      }
+
+      function expectDnf(specIn, specOut) {
+        var fDnf = specToDnf(specIn);
+
+        if(specOut) {
+          expect(fDnf.toSpec({forceType: true})).toEqual(specOut);
+        }
+      }
+
+      it("should preserve a filter already in DNF", function() {
+        expectDnf({
+          _: "or"
+        }, {
+          _: "or"
+        });
+      });
+
+      it("should apply De Morgan Rule 1 - NOT over AND", function() {
+        expectDnf({
+          _: "not",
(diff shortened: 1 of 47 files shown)


impl/client/src/test/javascript/pentaho/type/filter/abstract.spec.js
```diff
@@ -166,5 +166,312 @@ define([
...
+      it("should convert an empty `or` to false", function() {
+
+        expectDnf({
+          _: "or"
+        }, {
+          _: "false"
+        });
+      });
+
+      it("should convert an empty `and` to false", function() {
```
nunodelisboa: The description of the test does not match the test :)

dcleao:
@nantunes The filter types are also immutable, as you say, so it is now totally safe to create caches and it is not needed anymore to create clones.

About the `isIn`, we've talked here on making it private, in the documentation, and placing a JS comment on the class explaining why.

dcleao:
@nantunes About using workers, that's a great idea, as long as we can make the emission of the event and the application of the selection be asynchronously separated — seems feasible. On the safeguards, those would go well along with workers, allowing the user to cancels the selection altogether.

dcleao:
Rebased, due to a failing unit test caused from a change in a dependency (CDF).

dcleao:
Added @carlosrusso's commit fixing aspects related to the default selection mode.

dcleao:
Added commit with Default CCC vizs' styles for selection and activity.

PR #975, opened 2017-03-20
[BACKLOG-15218] Fix to palette unit test and removing redundant configurations

davidmsantos90:
@pentaho/millenniumfalcon @diogofscmariano please review

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.views.conf.js
+++ b/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.views.conf.js
@@ -58,10 +58,7 @@ define([
             invalidDataMessage_visible: false,
 
             plotFrameVisible: false,
-            plotSizeMin: {width: 600, height: 300},
 
-            clickable:  false,
-            selectable: false,
             hoverable:  true,
 
             // Legend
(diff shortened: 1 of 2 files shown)


impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.views.conf.js
```diff
@@ -60,8 +60,6 @@ define([
             plotFrameVisible: false,
             plotSizeMin: {width: 600, height: 300},
```
diogofscmariano: @davidmsantos90 does this means that all charts should have this default size min?

davidmsantos90: yeah, but you can override this. also if we think that this shouldn't be applied by default because of, e.g. CDF, we can also move this to container specific configs

diogofscmariano: yes. I know i can override this, but knowing this is a platform wide config, defining sizes for the charts might be too strong

nantunes: Those are minimum sizes... So I'm ok with either approach (having global defaults or not).

diogofscmariano: I suggest moving this to the plugins that really need

nantunes: 👍

davidmsantos90:
@diogofscmariano it has been removed :+1:

PR #994, opened 2017-03-29
[BACKLOG-15553] Sunburst white slices’ border, category labels, no le…

dcleao:
…gend.

Part 1 of Sunburst fixes. Still missing, is the gradient color step of 15%.

@pentaho/millenniumfalcon, @diogofscmariano please review.

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
+++ b/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
@@ -60,9 +60,7 @@ define([
       {
         priority: -5,
         select: {
-          type: [
-            "pentaho/visual/models/heatGrid"
-          ]
+          type: "pentaho/visual/models/heatGrid"
         },
         apply: {
           props: {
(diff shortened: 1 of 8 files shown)


impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
```diff
@@ -781,7 +813,7 @@ define([
         }
       },
       {
-        priority: -2,
```
diogofscmariano: i did it here with -2 in a way we still have the -1 to use in the plugins. What makes sense here?

dcleao: We're actually giving a value of -5 for viz-author default styles, and -1 for apps, I think, so that would be even more appropriate.

dcleao: Will make it uniform in a subsequent commit.

PR #997, opened 2017-03-29
[Backlog 15602] Added French, German and Japanese locales for vizApi 3.0

davidmsantos90:

Related with https://github.com/pentaho/pentaho-platform-plugin-geo/pull/163

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/visual/models/i18n/model.properties
+++ b/impl/client/src/main/javascript/web/pentaho/visual/models/i18n/model.properties
@@ -2,7 +2,6 @@
 
 ## Abstract Chart
 abstract.label=Abstract Chart
-#abstract.description=
 
 abstract.props.rows.label=Rows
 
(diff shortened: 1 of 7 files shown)


impl/client/src/main/javascript/web/pentaho/visual/models/i18n/model.properties
```diff
@@ -2,7 +2,7 @@
 
 ## Abstract Chart
 abstract.label=Abstract Chart
-abstract.description=
+#abstract.description=
```
nantunes: Why comment and not remove?

davidmsantos90: Well for the abstracts, I say they could be removed. I only left it there so that people know it exists.
But for the other locales I commented the ones missing translations.

I will remove the abstract ones

nantunes: Just asking. If it makes sense leave them.

PR #999, opened 2017-03-30
[BACKLOG-15654] Removed interactivity of trend lines (and their dots).

carlosrusso:

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
+++ b/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
@@ -109,6 +109,10 @@ define([
       arrowVisible: false,
       opacity: 1
     },
+    trendDot_ibits: 0,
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
```diff
@@ -109,6 +109,10 @@ define([
       arrowVisible: false,
       opacity: 1
     },
+    trendDot_ibits: 0,
+    trendDot_imask: -1,
+    trendLine_ibits: 0,
+    trendLine_imask: -1,
```
dcleao: This would be enough:
```js
trendDot_ibits: 0,
trendDot_imask: "ShowsActivity",
trendLine_ibits: 0,
trendLine_imask: "ShowsActivity",
```
otherwise we're preventing display of the tooltip and of drilling-down.

PR #1022, opened 2017-04-11
[BACKLOG-15790] Show null values in Heatgrid and Bubble charts

nantunes:
@pentaho/millenniumfalcon, @diogofscmariano please review.

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
+++ b/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
@@ -369,13 +369,23 @@ define(function() {
           extension: {
             // Plot
             sizeAxisRatio: 1 / 5,
-            sizeAxisRatioTo: "height", // plot area client height
+            sizeAxisRatioTo: "minwidthheight",
             sizeAxisOriginIsZero: true,
 
             // . dot
+            dot_lineWidth: function() {
+              // show nullShape
+              var v = !this.panel.visualRoles.size.isBound() || this.getSize() != null ? 0 : this.delegate();
+              return this.finished(v);
+            },
+            dot_strokeStyle: fillStyle2,
             dot_shapeSize: function() {
               var v = this.panel.visualRoles.size.isBound() ? this.delegate() : (5 * 5);
               return this.finished(v);
+            },
+
+            dot_DefaultVisible: function() {
+              return this.finished(true);
             }
           }
         }
@@ -669,7 +679,12 @@ define(function() {
             yAxisSizeMax: 80, // shouldn't it be: maxHorizontalTextWidth ??
 
             // . dot
-            dot_lineWidth: function() { return this.finished(0); },
+            dot_lineWidth: function() {
+              // show nullShape
+              var v = this.getSize() != null ? 0 : this.delegate();
+              return this.finished(v);
+            },
+            dot_strokeStyle: fillStyle1,
             dot_shapeRadius: function() { return this.finished(this.delegate()); },
             dot_fillStyle: fillStyle1
           }


impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
```diff
@@ -669,7 +679,12 @@ define(function() {
             yAxisSizeMax: 80, // shouldn't it be: maxHorizontalTextWidth ??
 
             // . dot
-            dot_lineWidth: function() { return this.finished(0); },
+            dot_lineWidth: function() {
+              // show nullShape
+              var v = this.getSize() != null ? 0 : this.delegate();
```
dcleao: Why not the same treatment of Bubble's `dot_lineWidth`?
I mean, either one has more code than it should or the other less than it should, no?

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
```diff
@@ -369,13 +369,23 @@ define(function() {
...
+              var v = !this.panel.visualRoles.size.isBound() || this.getSize() != null ? 0 : this.delegate();
+              return this.finished(v);
+            },
+            dot_strokeStyle: fillStyle2,
             dot_shapeSize: function() {
               var v = this.panel.visualRoles.size.isBound() ? this.delegate() : (5 * 5);
               return this.finished(v);
+            },
+
+            dot_DefaultVisible: function() {
```
dcleao: This extension point does not exist, does it?

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
```diff
@@ -369,13 +369,23 @@ define(function() {
           extension: {
             // Plot
             sizeAxisRatio: 1 / 5,
-            sizeAxisRatioTo: "height", // plot area client height
+            sizeAxisRatioTo: "minwidthheight",
```
dcleao: Just checking: UX wanted to change this long time decision?

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
nantunes: No, just something I forgot to delete. Was trying to show the null bubble even when there is no color bound (https://github.com/webdetails/ccc/blob/master/package-res/ccc/core/base/sign/dotSizeColor-sign.js#L53-L56).

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
nantunes: The null crosses at the top were being cut. That's the way I found to avoid https://github.com/webdetails/ccc/blob/master/package-res/ccc/plugin/scatter/scatter-plot-panel.js#L401-L407 and saw no side effects, but I was expecting your input on that.

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
nantunes: I know the above one is needed, as the bubble chart acts as scatter when the size isn't bound, and shows a border around the bubbles if the bound is not checked. It probably makes sense to be here also.

dcleao: The size visual role is optional in both charts. Not sure, but when size is not bound, `getSize()` probably returns `null`; that explains thee check for the size being bound in the Bubble chart, one that seems to also make sense in the HG.

PR #1023, opened 2017-04-13
[PIR-1189] Filter Type inconsistency

Advent51:
- The operator value wasn't set into the operatorSelect, that caused the setting of the first value from the list.


Advent51:
@pamval @mbatchelor please review

PR #1026, opened 2017-04-13
[BACKLOG-15788] Setting visualizations legend dot size to 8px

davidmsantos90:

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
+++ b/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
@@ -1405,7 +1405,7 @@ define([
         }
 
         // Set, before, whenever lineWidth is defined.
-        var dotSize = 16;
+        var dotSize = 8;
         var dotRadius = 4;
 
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
```diff
@@ -1405,7 +1405,7 @@ define([
         }
 
         // Set, before, whenever lineWidth is defined.
-        var dotSize = 16;
+        var dotSize = 8;
```
dcleao: dotSize = dotRadius * dotRadius = 4 * 4 = 16

dcleao: Also, the following needs to be in sync with these: 
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js#L778

Maybe a note should also be added on both sides.

PR #1031, opened 2017-04-19
[BACKLOG-14318] Update template

davidmsantos90:
@pentaho/millenniumfalcon @pentaho/bobafett please review.

Already tested in CDF project and this changes don't break the type parsing in the project.

The first commit was the template port from CDF, the second has the changes from CDF in the type.tmpl and the upgrade needed because of Common-ui


impl/client/src/doc/javascript/template/vizApi/tmpl/type.tmpl
```diff
@@ -1,7 +1,207 @@
 <?js
-    var data = obj;
-    var self = this;
-    data.forEach(function(name, i) { ?>
-<span class="param-type"><?js= self.linkto(name, self.htmlsafe(name)) ?></span>
-<?js if (i < data.length-1) { ?>|<?js } ?>
-<?js }); ?>
\ No newline at end of file
+  var data = obj;
```
davidmsantos90: This move was done in the CDF template (or the templates weren't in sync like they should), I wasn't aware that types were removed. I'm going to check this.

impl/client/src/doc/javascript/template/vizApi/tmpl/type.tmpl
```diff
@@ -1,7 +1,207 @@
...
+
+  var modifiers = "[?!]?";
+  var left_p = "\\(?";
+  var right_p = "\\)?";
+  var remainder = "$|(?:[|,]\\s*)(.+)";
+  var complexProps = "[<(?!\\w|,. )>]+";
+  var BACKBONE_TYPE_REGX = /backbone\.([a-z]+)/i;
+
+  // regx: /([?!]?)([\w.*]+)(?:())?/
+  var SIMPLE_TYPE_REGX = new RegExp("(" + modifiers + ")([\\w.*]+)(?:\\(\\))?");
```
davidmsantos90: The `*` matches types declared like `{*}`

Regarding the functions, from the data that was being parsed inside the type.tmpl file, this complex cases didn't reach this point. the most complex that would get to this point is `[?!]function[()]`
so I'm assuming the template is parsing this cases before sending them here.

impl/client/src/doc/javascript/template/vizApi/tmpl/type.tmpl
```diff
@@ -1,7 +1,207 @@
...
+
+  // regx: /^(?:([?!]?)([\w.]+))\.<([<(?!\w|,. )>]+)>/
+  var COMPLEX_TYPE_REGX = new RegExp("^(?:(" + modifiers + ")([\\w.]+))\\.<(" + complexProps + ")>");
```
davidmsantos90: This naming is being used here because of previous naming used in the template in CDF, and it is also used in jsdoc3 code. This way we keep it uniform in this context

davidmsantos90:

last commit depends on this PR -> https://github.com/pentaho/maven-parent-poms/pull/9

impl/client/src/doc/javascript/template/vizApi/publish.js
```diff
@@ -731,36 +670,11 @@ exports.publish = function(taffyData, opts, tutorials) {
...
-
-                  doclet.summary = summary;
-                  //console.log("summary: " + summary);
-                  //console.log("  description: " + desc);
-                  //console.log(" ");
-                } else {
-                  console.warn("Could not determine summary for: " + desc);
-                }
+                //Try to split when a "." or a ".</htmlTag>" is found.
+                //TODO: When markdown is present it fails the split and dumps all description in the summary.
```
dcleao: Don't understand. What remains to be done?
The first "." or `</htmlTag>` should be used to detect the end of the first phrase.
Which cases are successfully handled by the new code? And which fail?
Logging of cases in which the first phrase was not detected was a useful feature.

impl/client/src/doc/javascript/template/vizApi/tmpl/type.tmpl
dcleao: Still can't find all the types in the original `registerStandardJsTypes`.
Also, still this whole code is still in a tmpl file...

impl/client/src/doc/javascript/template/vizApi/tmpl/type.tmpl
dcleao: You still don't provide any examples of the several cases the regexes should match. This will be invaluable in the future.

PR #1034, opened 2017-04-25
[PRD-5879] Datepicker doesn't display a user selection if a date has a 30/45 minute time zone

Advent51:
- Offset formatting was fixed


Advent51:
@annapopova1 @tmorgner @rfellows @pamval @mbatchelor please review

tmorgner:
It might be a good idea to add a unit-test for that method. Writing the test is trivial, as you dont even need mocks - luckily that method uses only built-in functions. This will ensure that later changes or refactorings do not break your fix.

PR #1041, opened 2017-05-03
[BISERVER-12801] Change the message shown when deleting a data source.

joaormf:


impl/client/src/main/javascript/web/dataapi/nls/messages_de.properties
```diff
@@ -15,6 +15,7 @@ datasourcelistlbl_content=Datenquellen:
 editdatasourceimg_title=Die ausgew\u00e4hlte Datenquelle bearbeiten
 # deletedatasourceimg_title is the tooltip for the 'delete datasource' button on the datasource selection dialog
 deletedatasourceimg_title=Die ausgew\u00e4hlte Datenquelle l\u00f6schen
+DeleteDatasourceWarningTitle=
```
tmorgner: I doubt that you want an empty string as your translation. Those empty properties should be removed so that the translation can be looked up in the parent bundle.

(And yeah, be prepared to fix a few of these problems when weaving in newly translated strings. Its even more fun once you get to decide whether to encode or not to encode strings with entities based on where and how the string is used.)

joaormf:
I've removed the empty strings... please also note that the ja, fr and de messages weren't changed. If it's supposed to be changed on this context (and not by some other translating process or something) let me know.

tmorgner:
At some point during the release process we hand off the translations to a professional translation service and they (some time later) drop of a new completed set of translations to be merged in. When that happens you'll notice joy and excitement on everyone's face. I would expect the next drop somewhere closer to the 8.0 release date.

PR #1042, opened 2017-05-04
Make sure that the layout and properties panels retain their current open state

joaormf:


mbatchelor:
Quick question - shouldn't the _resizeGroup: function(group) method also have handling/understanding of the previousGroupsOpenState? Isn't that called when a group collapses or opens or closes?

joaormf:
@mbatchelor I believe it's not needed.

The previous open state is needed because whenever the __setConfiguration_ function is called the current _groups_ variable with the current TitlePanel instances is discarded. They get rebuit in the __createGroup_ function, without knowing if they were open or closed.

So my fix is to keep the previous state when the TitlePanel are destroyed and use them, if available, when a new TitlePanel is built.

pamval:
commit message should start with [ANALYZER-3557] - ... 

(assuming ANALYZER-3557 is the case this commit is for)

joaormf:
@pamval should I delete the PR/branch and create a new one, with the correct commit message?

pamval:
you don't need to do that, just amend the commit with the correct message.

PR #1043, opened 2017-05-05
[BACKLOG-14320] Added AbstractDataTable#filterRow method.

dcleao:
Without this method, one cannot easily know if filter includes a specific row.
Internal solutions use the `AbstractDataTable#filter` method, which is not useful in row by row scenarios or the private `pentaho/data/_ElementMock` wrapper.

The Sample documentation and third-party implementations need this to implement selection!

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/data/_AbstractTable.js
+++ b/impl/client/src/main/javascript/web/pentaho/data/_AbstractTable.js
@@ -247,7 +247,7 @@ define([
      * Returns a view for the subset of rows of a data table
      * that are selected by this filter.
      *
-     * @param {!pentaho.data.Table} dataTable - The data table to filter.
+     * @param {!pentaho.type.filter.Abstract} filter - The filter.
      *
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/pentaho/data/_AbstractTable.js
```diff
@@ -276,6 +276,25 @@ define([
...
+     *
+     * @param {!pentaho.type.filter.Abstract} filter - The filter.
+     * @param {number} rowIndex The row index (zero-based).
+     *
+     * @return {boolean} `true` if the filter includes the row; `false`, otherwise.
+     *
+     * @throws {pentaho.type.ValidationError} When the filter is not valid,
+     * the first error returned by the `validate` method.
+     */
+    filterRow: function(filter, rowIndex) {
```
nantunes: The method name isn't clear... How about `filterMatchesRow`?

dcleao:
Recreated here: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1045

PR #1055, opened 2017-05-18
[BACKLOG-16402] Doc reviews

davidmsantos90:
@pentaho/millenniumfalcon changes related with the spike to implement doc reviews

### DIFF
--- a/impl/client/src/main/doc-js/pentaho/IRange.jsdoc
+++ b/impl/client/src/main/doc-js/pentaho/IRange.jsdoc
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2016 Pentaho Corporation. All rights reserved.
+ * Copyright 2010 - 2017 Pentaho Corporation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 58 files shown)


impl/client/src/main/doc-js/pentaho/IRange.jsdoc
```diff
@@ -15,21 +15,20 @@
  */
 
 /**
- * A range represents an interval on an underlying, base, ordinal domain.
+ * A range represents an interval on an underlying, ordinal domain.
```
nantunes: That comma doesn't make sense if the apposition is gone.

impl/client/src/main/doc-js/pentaho/context/IContext.jsdoc
```diff
@@ -15,14 +15,14 @@
  */
 
 /**
- * The `IContext` interface exposes contextual information of the Pentaho Web Client Platform,
- * like, for example, its
+ * The `IContext` interface allows access to the Pentaho Web Client Platform.
```
davidmsantos90: @pentaho/millenniumfalcon I'm not sure with the change from exposes to allows access, what is your opinion?

dcleao: "Allows access" would be ok, if "to contextual information" hadn't been removed — that changes the meaning.

davidmsantos90: Thanks for the feedback :)
this was an attempt of mine, to make this uniform with doc suggestion in the other two files, I will change this after feedback from the doc team to a suggestion I gave taking into account what you said

PR #1061, opened 2017-05-23
[SP-3561][ANALYZER-3557] - Make sure that the layout and properties p…

mchen-len-son:
…anels retain their current open/close state


mchen-len-son:
@mbatchelor @pamval please take a look.
Thanks.

mchen-len-son:
@pamval can you please take a look?
thanks.

PR #1066, opened 2017-05-25
[SP-3480] Backport of PIR-1189 - Filter Type inconsistency (6.1 Suite)

Advent51:


Advent51:
@mbatchelor @pamval @mchen-len-son please review

PR #1080, opened 2017-06-02
[CLEANUP] Small fixes to JsDocs text. To be merged to 7.1-Doc as well…

dcleao:
… when possible.


impl/client/src/main/javascript/web/pentaho/type/_type.js
```diff
@@ -887,7 +887,7 @@ define([
       /**
        * Gets the style classes of this and any base types.
        *
-       * Does **not** modify the returned array.
+       * Do **not** modify the returned array.
```
carlosrusso: What about "The returned array should not be modified." ?

dcleao:
@carlosrusso please re-review.

PR #1083, opened 2017-06-09
[PIR-1339] Datepicker in PIR prompt incorrectly applies date format with minutes

Advent51:
- date format is used according to the dojo/date/locale


Advent51:
@mbatchelor @pamval @tmorgner @mchen-len-son please review

Please note that the https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/603 should be merged simultaneously.

PR #1086, opened 2017-06-09
[BACKLOG-16602] Refactor refinements

dcleao:

Merge before and followed by:
- DET: https://github.com/pentaho/pentaho-det/pull/870
- GEO: https://github.com/pentaho/pentaho-platform-plugin-geo/pull/188
- Analyzer: https://github.com/pentaho/pentaho-analyzer/pull/1540


impl/client/src/main/javascript/web/pentaho/type/element.js
```diff
@@ -97,27 +97,51 @@ define([
...
+         * @return {number} `-1` if `va` is considered _before_ `vb`; `1` is `va` is considered _after_ `vb`;
+         * `0`, otherwise.
+         */
+        compare: function(va, vb) {
+          // Quick bailout tests
+          if(va === vb) return 0;
+          if(va == null) return vb == null ? 0 : -1;
+          if(vb == null) return 1;
+
+          return this._areEqual(va, vb) ? 0 : this._compare(va, vb);
```
nantunes: Don't we expect `_compare` to return `0` when the values are equal?

impl/client/src/main/javascript/web/pentaho/type/list.js
```diff
@@ -208,6 +208,53 @@ define([
         return clone;
       },
 
+      // region configuration
+
+      /**
+       * Configures a list instance with a given configuration.
+       *
+       * When `config` is another list, behaviour is currently undefined.
```
nantunes: I would say that it is not implemented and that it throws an error. It is undefined in our heads (thus not implemented) but not in the code. This can make someone to think it is "ok" to pass a list, you just don't know what will happen.

impl/client/src/main/javascript/web/pentaho/visual/models/types/lineWidth.js
```diff
@@ -22,14 +22,13 @@ define([
 
   return function(context) {
 
-    var Refinement = context.get("pentaho/type/refinement");
+    var PentahoString = context.get("string");
```
nantunes: No longer numbers?

impl/client/src/main/javascript/web/pentaho/visual/samples/calc/model.js
```diff
@@ -57,9 +57,9 @@ define([
           {
             name: "operation",
             type: {
-              base: "refinement",
-              of:   "string",
-              facets: ["DiscreteDomain"],
+              base: "string",
```
nantunes: In the other models you omitted `base: "string"` because it is the default (right?).
You kept it here because it is the sample?

impl/client/src/main/javascript/web/pentaho/visual/models/types/lineWidth.js
dcleao: Copy & Paste...

impl/client/src/main/javascript/web/pentaho/visual/samples/calc/model.js
dcleao: When using specifications, the default value type of a **root** property is "complex", so here, we need to specify "number" to override the default. On inherited properties, the default value type is the base property's type.
In the other models, like in the `.../models/types/colorSet`? In these the base type, `"string"` is written in code: `PentahoString.extend( ... );`

impl/client/src/main/javascript/web/pentaho/type/element.js
dcleao: `_compare` is explicitly defined to receive values that are not equal. It's name could be `compareNonEqualNonNully` :-) .

PR #1091, opened 2017-06-21
[BISERVER-13677]: Report with Post Processing Formula that references…

TatsianaKasiankova:
… another parameter goes into a loop and it never stops

timeutil.js knows nothing about Daylight Savings Time and returns incorrect time offset when DST is in effect.
Fixed by using moment-timezone JS with DST data.


TatsianaKasiankova:
@mbatchelor @pamval @tmorgner @mchen-len-son please review.
Please merge it after https://github.com/webdetails/cdf/pull/931

PR #1101, opened 2017-07-06
[SP-3578][PIR-1334] [PIR-1327] - Fix z-index issues with date picker …

mchen-len-son:
…month lists.


mchen-len-son:
@mbatchelor @pamval @stepanovdg can you please take a look?

PR #1109, opened 2017-07-12
[PRD-5909]-Using the a period (.) in parameter name makes this parameter

PashkovichAndrei:
selection inworkable


PashkovichAndrei:
@tmorgner, could you please do review?

PR #1113, opened 2017-07-13
[BACKLOG-17504] DET: Geo Map doesn't render

dcleao:
@pentaho/millenniumfalcon, @afrjorge please review.


impl/client/src/main/javascript/web/jquery/jquery.conflict.js
```diff
@@ -0,0 +1,27 @@
+/*!
+ * Copyright 2010 - 2017 Pentaho Corporation. All rights reserved.
```
afrjorge: Since this is a new file, I think it should be `Copyright 2017`.

PR #1115, opened 2017-07-14
[BACKLOG-17688] - Change the prompt API to accept a parameter to enab…

pentaho-arua:
…le the "Select2" library in the CDF Select component


pentaho-arua:

PR #1119, opened 2017-07-17
[BACKLOG-16990] Added new filters to support GT, LT, GTE and LTE

nunodelisboa:
Please review @pentaho/bobafett @pentaho/millenniumfalcon

### DIFF
--- a/assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
+++ b/assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
+ * Copyright 2010 - 2017 Pentaho Corporation.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -92,6 +92,10 @@
   requireTypeInfo["pentaho/data/filter/property"] = {base: "pentaho/data/filter/abstract"};
   requireTypeInfo["pentaho/data/filter/isEqual"] = {alias: "=", base: "pentaho/data/filter/property"};
   requireTypeInfo["pentaho/data/filter/isIn"] = {alias: "in", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isGreater"] = {alias: ">", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isGreaterOrEqual"] = {alias: ">=", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isLess"] = {alias: "<", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isLessOrEqual"] = {alias: "<=", base: "pentaho/data/filter/property"};
 
   requireTypeInfo["pentaho/visual/base"] = {base: "model"};
   requireTypeInfo["pentaho/visual/base/view"] = {
--- a/impl/client/src/main/doc-js/pentaho/data/filter/_namespace.jsdoc
+++ b/impl/client/src/main/doc-js/pentaho/data/filter/_namespace.jsdoc
@@ -32,6 +32,10 @@
  *   * [pentaho/data/filter/not]{@link pentaho.data.filter.Not}
  *   * [pentaho/data/filter/property]{@link pentaho.data.filter.Property}
  *     * [pentaho/data/filter/isEqual]{@link pentaho.data.filter.IsEqual}
+ *     * [pentaho/data/filter/isGreater]{@link pentaho.data.filter.IsGreater}
+ *     * [pentaho/data/filter/isLess]{@link pentaho.data.filter.IsLess}
+ *     * [pentaho/data/filter/isGreaterOrEqual]{@link pentaho.data.filter.IsGreaterOrEqual}
+ *     * [pentaho/data/filter/isLessOrEqual]{@link pentaho.data.filter.IsLessOrEqual}
  *
  * @name pentaho.data.filter
  * @namespace
--- a/impl/client/src/main/doc-js/pentaho/data/filter/spec/IIsGreater.jsdoc
+++ b/impl/client/src/main/doc-js/pentaho/data/filter/spec/IIsGreater.jsdoc
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2017 Pentaho Corporation. All rights reserved.
+ * Copyright 2017 Pentaho Corporation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,16 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-define([
-  "./abstract"
-], function(abstractFactory) {
 
-  "use strict";
+/**
+ * The `spec.IIsGreater` interface describes the information associated with a is greater filter.
+ *
+ * @name pentaho.data.filter.spec.IIsGreater
+ *
+ * @interface
+ * @extends pentaho.data.filter.spec.IProperty
+ *
+ * @see pentaho.data.filter.IsGreater
+ */
 
-  return function(context) {
-
-    var Abstract = context.get(abstractFactory);
-
-    return Abstract._core.Property;
-  };
-});
+/**
+ * The expected value of the property.
+ *
+ * @name value
+ * @memberOf pentaho.data.filter.spec.IIsGreater#
+ * @type {!pentaho.type.UElement}
+ */
--- a/impl/client/src/main/javascript/web/pentaho/data/filter/KnownFilterKind.js
+++ b/impl/client/src/main/javascript/web/pentaho/data/filter/KnownFilterKind.js
@@ -59,6 +59,30 @@ define(function() {
     IsIn: "isIn",
 
     /**
+     * The [IsGreater]{@link pentaho.data.filter.IsGreater} filter kind.
+     * @default
+     */
+    IsGreater: "isGreater",
+
+    /**
+     * The [IsLess]{@link pentaho.data.filter.IsLess} filter kind.
+     * @default
+     */
+    IsLess: "isLess",
+
+    /**
+     * The [IsGreaterOrEqual]{@link pentaho.data.filter.IsGreaterOrEqual} filter kind.
+     * @default
+     */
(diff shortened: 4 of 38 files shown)


impl/client/src/main/javascript/web/pentaho/data/filter/standard.js
```diff
@@ -22,11 +22,15 @@ define([
...
+  "./isLess",
+  "./isGreaterOrEqual",
+  "./isLessOrEqual",
   "./true",
   "./false"
 ], function(
-    abstractFactory, treeFactory, propFactory, andFactory, orFactory, notFactory, isEqFactory, isInFactory,
-    trueFactory, falseFactory) {
+    abstractFactory, treeFactory, propFactory,
+    andFactory, orFactory, notFactory, isEqFactory, isInFactory, isGtFactory, isLtFactory, isGteFactory, isLteFactory, trueFactory, falseFactory) {
```
pentaho-arua: Can we have the same names for the factories everywhere? 
Since we already have `isGreaterFactory` in some files I would prefer to have it that way instead of `isGtFactory`.

nunodelisboa: This is used in this file alone and was named in line with the existing `isEqFactory`. The names used elsewhere do not use abbreviations. If you feel it is important I can rename all factories in this file, new and old, to not use abbreviations either.

pentaho-arua: Ok I can live with it :)

impl/client/src/main/doc-js/pentaho/data/filter/spec/IIsGreater.jsdoc
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @see pentaho.data.filter.IsGreater
+ */
+
+/**
+ * The expected value of the property.
+ *
+ * @name value
+ * @memberOf pentaho.data.filter.spec.IIsGreater#
+ * @type pentaho.type.UValue
```
dcleao: `@type {pentaho.type.UValue}`. Applies to other filter types as well.
Assuming `null` can be specified, although the resulting filter will not be valid.

impl/client/src/main/doc-js/pentaho/data/filter/spec/IIsGreater.jsdoc
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The `spec.IIsGreater` interface describes the information associated with an equality filter.
```
dcleao: "associated with an equality filter" ?
The same mistake is present in the other filter types.

impl/client/src/main/javascript/web/pentaho/data/filter/KnownFilterKind.js
```diff
@@ -59,6 +59,30 @@ define(function() {
     IsIn: "isIn",
 
     /**
+     * The [IsGreater]{@link pentaho.data.filter.isGreater} filter kind.
```
dcleao: `isGreater` -> `IsGreater`

impl/client/src/main/javascript/web/pentaho/data/filter/isGreater.js
```diff
@@ -0,0 +1,102 @@
...
+     */
+
+    return PropertyFilter.extend("pentaho.data.filter.IsGreater", /** @lends pentaho.data.filter.IsGreater# */{
+
+      /** @inheritDoc */
+      get kind() {
+        return KnownFilterKind.IsGreater;
+      },
+
+      // TODO: In the future, review if value should be of type pentaho.type.Value.
```
dcleao: I know this is like it is in IsEqual, but I don't think this applies anymore.
We should make these `value` properties be of type `pentaho.type.Element`. List values wouldn't be a good fit.
Please fix this in IsEqual as well.
`valueType: "element"`
`@type {pentaho.type.Element}`

impl/client/src/main/javascript/web/pentaho/data/filter/isGreater.js
```diff
@@ -0,0 +1,102 @@
...
+      /**
+       * Gets the expected value of the property.
+       *
+       * This getter is a shorthand for `this.getv("value")`.
+       *
+       * @type {any}
+       *
+       * @readOnly
+       */
+      get value() {
```
dcleao: Also, please remove the getter implementation, but leave a virtual doc there. The isReadOnly attribute in the declaration of the property is sufficient to achieve a similar effect (I think it still generates the setter, but throws when set).

impl/client/src/main/doc-js/pentaho/data/filter/spec/IIsGreater.jsdoc
nunodelisboa: `pentaho.type.USimple` solves that issue, right? by disallowing Nully values.

annapopova1:
Looks like we need additionally declare new types in requirejs config

annapopova1:
And one more possible problem is cycling dependencies. For example, isGreater depends on isLessOrEqual and isLessOrEqual depends on isGreater

nunodelisboa:
@annapopova1 You are right about the dependencies. @dcleao was kind enough to solve that problem for me.

assemblies/platform-plugin/src/main/resources/resources/web/common-ui-require-js-cfg.js
```diff
@@ -93,6 +93,10 @@
   requireTypeInfo["pentaho/data/filter/property"] = {base: "pentaho/data/filter/abstract"};
   requireTypeInfo["pentaho/data/filter/isEqual"] = {alias: "=", base: "pentaho/data/filter/property"};
   requireTypeInfo["pentaho/data/filter/isIn"] = {alias: "in", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isGreater"] = {alias: ">", base: "pentaho/data/filter/property"};
```
annapopova1: don't forget change year in copyright

impl/client/src/test/javascript/pentaho/data/filter/isGreater.spec.js
```diff
@@ -0,0 +1,118 @@
...
+define([
+  "pentaho/data/filter/isGreater",
+  "pentaho/type/Context",
+  "pentaho/type/complex",
+  "./propertyUtils"
+], function(isGreaterFactory, Context, complexFactory, propertyUtils) {
+
+  "use strict";
+
+  /* global describe:true, it:true, expect:true, beforeEach:true*/
```
annapopova1: Could you explain the reason of usage this comment?

dcleao: I suspect that Nuno just continued the practice in other files, so I'll try to justify. This helps IDEA to know that these variables are valid globals. It would be more correct to say:
```js
/* global describe:false, it:false, expect:false, beforeEach:false*/
```
Meaning that the variables can only be read from.

Also, in these days, I prefer the shorter form:
```js
/* globals describe, it, expect, beforeEach*/
```

Anyway, probably there is some editor setting that would achieve the same globally. Is that why you ask?

If this setting could be globally set in an `eslint.rc` file or similar, that would then take effect with proper editor configuration, it would be a better alternative.
Pure editor solutions, on the other hand, are not as good, as these are not stored in the source control system.

dcleao:
@annapopova1 
> should we describe new filter types in _namespace.jsdoc file similar isEqual filter or it's not necessary?

Definitely, yes! Well spotted.

impl/client/src/test/javascript/pentaho/data/filter/isGreater.spec.js
annapopova1: I asked because it looks like to linter rules. Usually declaring `jasmine` env to `.eslintrc.json` is enough. I use VS code with linter plugin and it works fine for me. I don't use IDEA but I suppose so powerful tool should have something similar... because for now it looks really strange to declare this comment in each test source js files.

nunodelisboa:
@dcleao Modifications were made to make sure toDnf() and andNot() work with the new filters. New unit tests were added to specifically target those use cases.

impl/client/src/test/javascript/pentaho/data/filter/isGreater.spec.js
dcleao: Like many times, the reasons are historical. Definitely something we can improve. Because this is now widespread through all pentaho/* testing code, we'd probably better do this cleanup in one shot, within a tech-debt story. Do you agree?

annapopova1: I agree to clean up old code (existing files) according to future a tech-debt story. But for new files I prefer to not copy past this from old examples.

impl/client/src/main/doc-js/pentaho/data/filter/_namespace.jsdoc
```diff
@@ -32,6 +32,10 @@
  *   * [pentaho/data/filter/not]{@link pentaho.data.filter.Not}
  *   * [pentaho/data/filter/property]{@link pentaho.data.filter.Property}
  *     * [pentaho/data/filter/isEqual]{@link pentaho.data.filter.IsEqual}
+ *     * [pentaho/data/filter/isEqual]{@link pentaho.data.filter.IsGreater}
```
dcleao: Copy&paste, copy&paste, copy&paste ;-)

impl/client/src/test/javascript/pentaho/data/filter/abstract.spec.js
```diff
@@ -176,302 +177,475 @@ define([
...
-          _: "and",
-          o: [
-            {_: "=", p: "a", v: 1},
-            {
-              _: "not",
-              o: {
-                _: "or",
+        });
+
+        it("correctly deals with terminal filters other than isEquals", function(){
```
dcleao: I think you are missing a test for the case where it is `b` that has an inequality operator. I also think that that case is currently unhandled in the code.

dcleao: Now I see my confusion. The case is being handled and I got confused with the `a`, `b` property names of the example...

impl/client/src/main/doc-js/pentaho/data/filter/_namespace.jsdoc
nunodelisboa: yank and paste :) vi-fan here

PR #1120, opened 2017-07-17
[BACKLOG-17154] In Analyzer, the legend bullets do not appear when th…

dcleao:
…e bullet style in Line chart is changed to cross

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
+++ b/impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
@@ -1411,16 +1411,18 @@ define([
         }
 
         // Set, before, whenever lineWidth is defined.
-        var dotSize = 16;
+        // var dotSize = 16;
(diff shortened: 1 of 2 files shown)


impl/client/src/main/javascript/web/pentaho/ccc/visual/abstract.js
```diff
@@ -1411,16 +1411,18 @@ define([
         }
 
         // Set, before, whenever lineWidth is defined.
-        var dotSize = 16;
+        // var dotSize = 16;
```
nantunes: Why comment instead of remove? Not sure if it will be needed?

dcleao: I thought only of providing some form of guidance to the developer. The _effective_ area (size) of the circle will still be `4 * 4 = 16`. However, one part of the radius will be given by the "circle" border and the other part by the "circle" fill.

PR #1126, opened 2017-07-20
[SP-3665] Backport of PIR-1339 - Datepicker in PIR prompt incorrectly applies date format with minutes (6.1 Suite)

Advent51:


Advent51:
@pamval @tmorgner @mchen-len-son please review

PR #1142, opened 2017-08-07
[BACKLOG-18124] Created isLike filter for affix matching in the forma…

nunodelisboa:
…tted value

Please review @pentaho/millenniumfalcon @pentaho/bobafett

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
+++ b/impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
@@ -1,0 +1,128 @@
+/*!
+ * Copyright 2017 Pentaho Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define([
+  "../KnownFilterKind"
+], function(KnownFilterKind) {
+
+  "use strict";
+
+  return function(filter) {
+
+    /**
+     * @name pentaho.data.filter.IsLike.Type
+     * @class
+     * @extends pentaho.data.filter.Property.Type
+     *
+     * @classDesc The type class of the is-like filter type.
+     *
+     * For more information see {@link pentaho.data.filter.IsLike}.
+     */
+
+    /**
+     * @name pentaho.data.filter.IsLike
+     * @class
+     * @extends pentaho.data.filter.Property
+     * @amd {pentaho.type.Factory<pentaho.data.filter.IsLike>} pentaho/data/filter/isLike
+     *
+     * @classDesc The `IsLike` class represents a simple string matching filter.
+     * This filter selects elements having the formatted value match an affix
+     * (ie prefix, suffix or infix) of a reference value: [value]{@link pentaho.data.filter.IsLike#value}.
+     *
+     * @description Creates an is-like filter instance.
+     *
+     * @constructor
+     * @param {pentaho.data.filter.spec.IIsLike} [spec] - An is-like filter specification.
+     */
+
+    filter.IsLike = filter.Property.extend(/** @lends pentaho.data.filter.IsLike# */{
+
+      /** @inheritDoc */
+      get kind() {
+        return KnownFilterKind.IsLike;
+      },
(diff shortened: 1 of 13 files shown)


impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * @name pentaho.data.filter.IsLike
+     * @class
+     * @extends pentaho.data.filter.Property
+     * @amd {pentaho.type.Factory<pentaho.data.filter.IsLike>} pentaho/data/filter/isLike
+     *
+     * @classDesc The `IsLike` class represents a simple string matching filter.
+     * This filter selects elements having the formatted value match an affix
+     * (ie prefix, suffix or infix) of a reference value: [value]{@link pentaho.data.filter.IsLike#value}.
```
annapopova1: just interesting how it works with infix. Do we really support it?

impl/client/src/test/javascript/pentaho/data/filter/isLike.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      it("should return '(like  valueKey false false)' when no property is set'", function() {
+        var filter  = new IsLikeFilter({value: "str"});
+
+        expect(filter.contentKey).toBe("(like  str false false)");
+      });
+
+      it("should return '(like   false false)' when no property or value are set'", function() {
+        var filter  = new IsLikeFilter();
+
+        expect(filter.contentKey).toBe("(like   false false)");
```
annapopova1: Can contentKey contain several spaces? Is it correct?

impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
nunodelisboa: the infix case is the only one we directly support in DET :) prefix and suffix are supported only as the degenerate case of the infix. However, this is a general filter in common ui, there is no reason not to explicitly support all cases.

impl/client/src/test/javascript/pentaho/data/filter/isLike.spec.js
nunodelisboa: yes, missing values are represented as empty spaces, otherwise the content key of a filter with two (or more) optional parameters could not distinguish between having the first parameter empty and the second defined and vice-versa.

impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
```diff
@@ -0,0 +1,128 @@
...
+       *
+       * @name value
+       * @memberOf pentaho.data.filter.IsLike#
+       * @type {pentaho.type.Element}
+       *
+       * @readOnly
+       */
+
+      /** @inheritDoc */
+      _operation: function(value, formattedValue) {
```
dcleao: This function is not formatted according to the eslint rules of common-ui.

impl/client/src/main/javascript/web/pentaho/data/filter/_core/isLike.js
```diff
@@ -0,0 +1,128 @@
...
+          reference = value;
+        }
+
+        var firstOccurrence = -1;
+        if (reference != null){
+          firstOccurrence = reference.indexOf(caption);
+        }
+        return firstOccurrence > -1 &&
+          (!anchorStart || firstOccurrence === 0) &&
+          (!anchorEnd || reference.endsWith(caption));
```
dcleao: `String#endsWith` is not yet generally available. At least IE11 does not have it.

impl/client/src/main/javascript/web/pentaho/data/filter/_core/property.js
```diff
@@ -87,7 +88,7 @@ define(function() {
        */
       _contains: function(elem) {
         var prop = this.property;
-        return elem.$type.has(prop) && this._operation(elem.getv(prop));
+        return elem.$type.has(prop) && this._operation(elem.getv(prop), elem.getf(prop));
```
dcleao: Performance reasons would dictate that `_operation` would start receiving the element and would only get what it actually needs.

impl/client/src/test/config/javascript/require.config.js
```diff
@@ -94,6 +94,7 @@
   requireTypeInfo["pentaho/data/filter/isGreaterOrEqual"] = {alias: ">=", base: "pentaho/data/filter/property"};
   requireTypeInfo["pentaho/data/filter/isLess"] = {alias: "<", base: "pentaho/data/filter/property"};
   requireTypeInfo["pentaho/data/filter/isLessOrEqual"] = {alias: "<=", base: "pentaho/data/filter/property"};
+  requireTypeInfo["pentaho/data/filter/isLike"] = {alias: "like", base: "pentaho/data/filter/property"};
```
dcleao: This same line should be added to the runtime require configuration file.

PR #1148, opened 2017-08-21
[PIR-1355] Fields highlighting works incorrectly for multiple selection

mchen-len-son:
- reworked fields selection reset on mouse right click


mchen-len-son:
@pamval please take a look.
This is a fix for the regression in current patch.
thanks.

PR #1151, opened 2017-08-21
[SP-3679][PIR-1355] Fields highlighting works incorrectly for multipl…

mchen-len-son:
…e selection

 - reworked fields selection reset on mouse right click


mchen-len-son:
@pamval please have a look
thanks.

PR #1157, opened 2017-09-11
[BACKLOG-18432] I want to modify the color palette from VizAPI 3.0

dcleao:
Merge followed by:
* https://github.com/pentaho/eng-sandbox/pull/22
* https://github.com/pentaho/pentaho-engineering-samples/pull/32
* https://github.com/pentaho/pentaho-analyzer/pull/1612
* https://github.com/pentaho/pentaho-det/pull/1002
* https://github.com/pentaho/pentaho-platform-plugin-geo/pull/205
* https://github.com/webdetails/cdf/pull/953
* https://github.com/webdetails/cgg/pull/105
* https://github.com/webdetails/cde/pull/798
* https://github.com/pentaho/pentaho-platform/pull/3817
* https://github.com/pentaho/pentaho-det-ee/pull/496


docs/platform/visual/samples/bar-d3-sandbox/step3-view-creation.md
```diff
@@ -11,37 +11,34 @@ layout: default
...
-        id: module.id,
-        props: [
-          {
-            name: "model",
-            type: barModelFactory
-          }
-        ]
-      },
+  return [
+    "pentaho/visual/base/view",
```
nantunes: Are these AMD module IDs? If yes, why not include "d3" in here? If not, and they are exclusively type IDs (or expressions), then the syntax should become clearly different (like ```"pentaho.visual.samples.bar.Model"```).

Very confusing, altogether.

impl/client/src/main/javascript/web/pentaho/ccc/visual/barHorizontal.js
```diff
@@ -14,27 +14,27 @@
...
-  "./barAbstract"
-], function(module, modelFactory, baseViewFactory) {
+  "module"
+], function(module) {
 
   "use strict";
 
-  return function(context) {
+  return [
+    "pentaho/ccc/visual/barAbstract",
```
nantunes: Did we lost the ability of using relative namespaces?

impl/client/src/main/javascript/web/pentaho/type/Context.js
```diff
@@ -16,40 +16,278 @@
...
   "../util/error",
   "../util/object",
-  "../util/fun"
-], function(localRequire, module, service, typeInfo, bundle, standard, SpecificationContext, SpecificationScope,
-    mainPlatformEnv, configurationService,
+  "../util/fun",
+  "../debug",
+  "../debug/Levels",
+  "../util/logger",
+  "./standard" // so that r.js sees otherwise invisible dependencies.
```
nantunes: Can't this be handled in r.js configuration?

docs/platform/visual/samples/bar-d3-sandbox/step3-view-creation.md
nantunes: Any way to inject a reference to the context itself?

Are type aliases supported? Could I just inject ```"="```?

nantunes: Also, forgot to ask: why don't type factories receive the config and do whatever they want with it, like instances?

impl/client/src/main/javascript/web/pentaho/visual/color/paletteProperty.js
```diff
@@ -0,0 +1,218 @@
...
+  "pentaho/util/error"
+], function(module, bundle, ValidationError, typeUtil, O, error) {
+
+  "use strict";
+
+  return [
+    "property",
+    "pentaho/visual/color/level",
+    "pentaho/visual/color/palette",
+    {$instance: {type: ["pentaho/visual/color/palette"]}},
```
nantunes: Why is `allPalettesList` needed? Is it to ensure all palettes are loaded?

docs/platform/visual/samples/bar-d3-sandbox/step1-environment-preparation.md
```diff
@@ -78,46 +78,51 @@ npm install
...
-            // 2. Set the DOM container dimensions.
-            viewSpec.domContainer.style.width = viewSpec.width + "px";
-            viewSpec.domContainer.style.height = viewSpec.height + "px";
-        
-            var BaseView = context.get(baseViewFactory);
-            BaseView.createAsync(viewSpec).then(function(view) {
-              // Render the visualization.
-              view.update();
-            });
+            Context.createAsync({application: "viz-api-sandbox"})
```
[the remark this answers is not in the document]
dcleao: Yes, in fact.

docs/platform/visual/samples/bar-d3-sandbox/step3-view-creation.md
dcleao: 1. Yes, they're AMD module IDs, which, at the moment, are the identifiers of Type API types and instances. Only Type API type dependencies are included here. Other regular AMD dependencies are left where they were.

impl/client/src/test/javascript/pentaho/visual/base/view.spec.js
```diff
@@ -819,7 +819,7 @@ define([
         // _updateSize is still updating
 
         // Change the view's selection
-        view.selectionFilter = null;
+        view.selectionFilter = {_: "or"};
```
nantunes: Why the change? Setting to `null` doesn't trigger the change detection?

impl/client/src/main/javascript/web/pentaho/ccc/visual/barHorizontal.js
dcleao: In the initial development of the "module dependency id" concept yes. Later, I had to add that ability back, to support relative ids to sibling modules in OSGI projects, where we don't know the base module id.

impl/client/src/main/javascript/web/pentaho/type/Context.js
dcleao: In theory yes. In practice, it would be much less practical. It is Analyzer that is doing the r.js build. I just think that it doesn't kill us to, in some cases, *also* declare the dependencies in the AMD module directly.

impl/client/src/main/javascript/web/pentaho/visual/color/paletteProperty.js
dcleao: Yep, this is a new module dependency syntax for loading instances. Because it is a list type that is being requested, all registered instance palettes are pre-loaded. The class code below assumes that every registered palette can be found synchronously.

impl/client/src/test/javascript/pentaho/visual/base/view.spec.js
dcleao: Setting to null triggers obtaining a new default value. However the default value semantics now has changed. Before we'd always clone the default value. Now, if that semantics is desired, a default value function must be supplied that creates a new instance each time.
The selectionFilter default value wasn't changed — as it is an instance of an immutable class — setting to `null` no longer yields a new value and, as such, does not trigger an on change anymore.

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
```diff
@@ -44,6 +44,31 @@ define(function() {
...
     rules: [
+      // region Viz API Standard Color Palettes Context Registration
+      {
+        priority: RULE_PRIO_VIZ_DEFAULT,
+        select: {
+          type: "pentaho/type/Context"
+        },
+        apply: {
+          instances: {
+            "pentaho/visual/color/palettes/nominalPrimary":    {type: "pentaho/visual/color/palette", priority: -10},
```
nantunes: A totally different way of registering stuff. We have `pentaho/service`, `pentaho/typeInfo` and by Context configuration.

impl/client/src/main/javascript/web/pentaho/type/Context.js
```diff
@@ -16,40 +16,278 @@
...
+   * @property {!Array.<string>} deps - The type's dependencies. Set when the AMD module is loaded.
+   *
+   * @property {Class.<pentaho.type.Instance>} Ctor - The type's constructor.
+   * Only not `null` when the type has already been created.
+   *
+   * @property {Promise.<Class.<pentaho.type.Instance>>} promise - A promise for the type's constructor.
+   * Only not `null` since the type has begun loading.
+   */
+
+  var TypeHolder = Base.extend({
```
nantunes: Isn't `TypeHolder` similar to `InstancesContainer`? Maybe move it to its own file?

nantunes: Ok, was seeing it wrong... `InstanceHolder` is a inner class of `InstancesContainer`, as `TypeHolder` is a inner class of `Context`.

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
dcleao: Yes.

1. `pentaho/service` is a low level inversion of control mechanism for AMD modules
2. `pentaho/typeInfo` declares types and works in the realm of types, not AMD modules; it happens that, currently, the ids of types are the ids of their AMD modules. Loading types by base type, in the Context, requires only that we know the relationship between types; where a type lives is a secondary deal; in our case, by convention, the type lives in a module with the same id as the type and type modules have a known value type. The metadata about types does not change with the environment.
3. Instances are yet another different kind of thing. We could argue that an instance does not change its type depending on the environment. So, it could have an environment independent way of stating its type. The priority however is inevitably tied to the environment and is tied to the configuration of the Context.

I also remember discussing the Context configuration solution with you. The part about changing to use `pentaho/typeInfo` instead of `pentaho/service`, that was mentioned several times, as a future thing I'd like to change, but we never discussed its pros and cons.

So, I do believe different kinds of problems require different solutions. For me, this is an instance of that general rule.

impl/client/src/main/javascript/web/pentaho/type/Context.js
dcleao: I believe that you mean that `TypeHolder` is similar to `InstanceHolder`. Yes they are very similar, but are also different in several places. Creating a separate base class would be an additional effort for me, specially taking into account the time it already took to develop all this. On a second stride, we would do that abstraction.

docs/platform/visual/samples/bar-d3-sandbox/step3-view-creation.md
graimundo: For the record. Like we talked, in the future we should revisit the way that we are declaring dependencies in multiple ways and check if we can reduce them (to whichever is better).  

I personally would also like to revisit the way (syntax) that we ask for multiple instances of a given type.

impl/client/src/main/doc-js/pentaho/type/spec/IInstanceResolve.jsdoc
```diff
@@ -0,0 +1,111 @@
...
+ * can also be specified.
+ *
+ * ##### Resolve instance by list type syntax
+ *
+ * Denotes the list of all of the registered instances of
+ * the given list type's [element type]{@link pentaho.type.List.Type#of},
+ * ordered from highest to lowest [priority]{@link pentaho.type.spec.IInstanceDeclaration#priority}.
+ *
+ * ```json
+ * {$instance: {type: ["my/Service"]}}
```
graimundo: Like I mentioned I would prefer to have an alternative syntax to represent getting the list of all registered instances of a given type. For later discussion on possible alternatives or if the current representation is the one we want to stick with.

impl/client/src/main/doc-js/pentaho/type/spec/IInstanceResolve.jsdoc
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * An instance predicate function that restricts the set of denoted instances.
+ *
+ * @name filter
+ * @memberOf pentaho.type.spec.IInstanceResolve#
+ * @type {function(!pentaho.type.Instance) : boolean}
+ */
+
+/**
+ * Indicates that there must be at least one instance being denoted.
```
graimundo: Do you mean that there must exist at least one registered instance? What's the behaviour when isRequired=true and no instance is found?

impl/client/src/main/doc-js/pentaho/type/spec/InstanceReservation.jsdoc
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+/**
+ * The `InstanceReservation` enumeration contains the possible reservation modes
+ * that a synchronous instance resolution operation can use.
+ *
+ * The synchronous instance resolution operations are:
+ * * [InstancesContainer#getById]{@link pentaho.type.InstancesContainer#getById}
+ * * [InstancesContainer#getByType]{@link pentaho.type.InstancesContainer#getByType}
+ * * [InstancesContainer#getAllByType]{@link pentaho.type.InstancesContainer#getByType}
```
graimundo: Typo. It seems that *getAllByType* its pointing to the wrong impl *getByType*

impl/client/src/main/doc-js/pentaho/type/spec/InstanceReservation.jsdoc
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The `InstanceReservation` enumeration contains the possible reservation modes
+ * that a synchronous instance resolution operation can use.
+ *
```
graimundo: Should we list the reservation modes here for quickly knowing the possible modes at the beginning of the doc?

impl/client/src/main/javascript/web/pentaho/type/instance.js
```diff
@@ -67,8 +67,8 @@ define([
      *   <code>greeting</code> and a method <code>doSomething</code>.
      * </caption>
      *
-     * require(["pentaho/type/Context"], function(Context) {
-     *   var context = new Context();
+     * require(["pentaho/type/context!"], function(context) {
```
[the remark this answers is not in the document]
dcleao: I intended to have developed such an AMD plugin... but ended up not doing so. Will fix the example.

impl/client/src/main/doc-js/pentaho/type/spec/IInstanceResolve.jsdoc
dcleao: Yes, but in the context in which this text is written, registration is unknown, so we simply refer to existence...

impl/client/src/main/doc-js/pentaho/type/spec/InstanceReservation.jsdoc
dcleao: This is a JsDocs enum. It is already formatted as a table of values.

impl/client/src/main/doc-js/pentaho/type/spec/IInstanceResolve.jsdoc
dcleao: In that case, an error is thrown when the specification is evaluated.

impl/client/src/main/javascript/web/pentaho/visual/config/vizApi.conf.js
graimundo: To revisit in the future the "ways of registering stuff".

PR #1189, opened 2017-11-04
[PIR-1311]-The japanese symbols displays like &#22823;

PashkovichAndrei:


PashkovichAndrei:
@dcleao could you please do review?

mchen-len-son:
@tmorgner @dcleao can you please take a look?

PR #1200, opened 2017-11-17
[PIR-1311]-fix for broken build

PashkovichAndrei:


PashkovichAndrei:
@dcleao @nantunes there is a fix for broken build after merging previous changes. could you please do review?

PR #1205, opened 2017-11-20
[PIR-1311] The japanese symbols displays like &#22823;

mchen-len-son:
- fixed function call context


mchen-len-son:
@mbatchelor @pamval please take a look. this is the cure for the failing tests in PIR
see https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/653

PR #1213, opened 2017-11-28
[BACKLOG-19949] Change the View#update operation to be implemented as…

dcleao:
… an action.

* Implemented `EventSource#_emitGenericAllAsync` - a missing piece of functionality for the execution of generic, async actions.
* Split `action.Base` into `action.Base` and `action.Execution`; this makes things much cleaner.
* Updated the existing actions, `pentaho.visual.action.Select` and `pentaho.visual.action.Update` to use the new split pattern.
* Moved the implementation of `ITarget` in `base.View` into a mixin class: `pentaho.type.action.impl.Target`.
* Updated/completed `base.View` and `EventSource` unit tests.
* Action et. al. unit tests are still to be done in a separate issue.

Merge with:
* https://github.com/pentaho/pentaho-analyzer/pull/1685
* https://github.com/pentaho/pentaho-det/pull/1074
* https://github.com/pentaho/pentaho-engineering-samples/pull/42


impl/client/src/main/javascript/web/pentaho/util/object.js
```diff
@@ -226,7 +226,8 @@ define(["./has"], function(has) {
      * @param {string} property - The name of property.
      * @param {Object} lcaExclude - A lowest-common-ancestor object whose inherited properties should
      * not be returned.
-     * @return {?Object} The [property descriptor]{@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty}.
+     * @return {?Object} The
+     * [property descriptor]{@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty}.
```
davidmsantos90: We should not be adding the complete link in the documentation, the template already has logic in place to handle most links to mdn global objects reference documentation.

But this isn't one of them, so we need to add support for it before simplifying the link.

PR #1214, opened 2017-11-28
[Backlog-19948] Remove aggregation / order by attributes from visual role mappings

davidmsantos90:

Removed both properties, updated docs and unit tests
and also did some smoke tests Analyzer and DET


impl/client/src/main/javascript/web/pentaho/visual/role/property.js
```diff
@@ -915,26 +910,13 @@ define([
...
-                } else {
-                  message = bundle.format(
-                    bundle.structured.errors.property.attributeDuplicate,
-                    {
-                      name: dataAttr,
-                      role: this
-                    });
-                }
+                var attributeDuplicateMessage = bundle.structured.errors.property.attributeDuplicate;
+                var erroMessage = bundle.format(attributeDuplicateMessage, {
```
nantunes: Should be `errorMessage`.

(this message failed to go with the review)

impl/client/src/main/javascript/web/pentaho/visual/role/mappingAttribute.js
```diff
@@ -144,15 +143,13 @@ define([
...
          * is [quantitative]{@link pentaho.visual.role.Level.Type#isQuantitative}.
          *
-         * This key is composed by the value of the [name]{@link pentaho.visual.role.MappingAttribute#name} and
-         * the [aggregation]{@link pentaho.visual.role.MappingAttribute#aggregation} properties.
+         * This key is composed by the value of the [name]{@link pentaho.visual.role.MappingAttribute#name} property.
          *
          * @type {string}
          * @readOnly
          */
         get keyQuantitative() {
```
dcleao: @davidmsantos90 The reason for the existence of `keyQualitative` and `keyQuantitative` in the first place was for the difference in uniqueness caused by the `aggregation` property.
Once this property has been removed, there is no need for two key properties. Maybe even there's no need to even create a getter for this and the user should just know that the key is the `name` property.

We cannot yet use `Complex#$key`, and override it to return the value of `name`, because `name` is not immutable and the list implementation (used by `Mapping#attributes`) is not prepared for keys to change along the way. So, we cannot improve things on this end, for now, and thus need to continue explicitly validating attribute `name` uniqueness in the attributes list...

impl/client/src/main/javascript/web/pentaho/visual/role/property.js
```diff
@@ -915,26 +910,13 @@ define([
               var key = isQuant ? roleAttr.keyQuantitative : roleAttr.keyQualitative;
```
dcleao: See, here, `key = roleAttr.name`. No need for `isQuant`, `keyQuantitative` and `keyQualitative`.

PR #1215, opened 2017-11-29
[PIR-1311]-fix for found regression

PashkovichAndrei:


PashkovichAndrei:
@nantunes @dcleao could you please have a look?

impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
```diff
@@ -420,7 +420,9 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
   _unescapeAmpIfExist: function(str) {
     if (Array.isArray(str)) {
       str.forEach(function(val) {
-        val.replace(/&amp;/g, '&');
+        if (typeof str === 'string' || str instanceof String) {
+          val.replace(/&amp;/g, '&');
```
dcleao: Unless `val` is actually a `String`, not common, replacing characters on a string returns a new string and thus have no side effect on the given array.
So I think you should do `str[index] = val.replace( ... );`, where `index` is the second argument to the `forEach`.

The same applies to the replace line below, when not in the forEach case.
Did this ever work? Are we using `String` objects — which is usually not good practice — for this to  work?

PashkovichAndrei:
@dcleao, this is a found regression: image
I add one more if statement into forEach method to avoid NPEs. Except check only for null, I've added check for String objects & string literals: `typeof str === 'string' || str instanceof String`; this check will be true only if we pass String object or string literal into it.

Also I've reworked PR regarding your comment above. Could you please do review?

impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
```diff
@@ -419,11 +419,13 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
 
   _unescapeAmpIfExist: function(str) {
     if (Array.isArray(str)) {
-      str.forEach(function(val) {
-        val.replace(/&amp;/g, '&');
+      str.forEach(function(val, index) {
+        if (typeof str === 'string' || str instanceof String) {
```
dcleao: Should it be: `if (typeof val === 'string' || val instanceof String)`
You should start testing your changes...

PashkovichAndrei: Sorry, my bad...

PR #1219, opened 2017-12-04
[BACKLOG-20128] Fixed amd i18n loader when using pentaho server on the default http(s) port

davidmsantos90:

This was happening because javascript URL omits the port value when it is the scheme default port

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/i18n.js
+++ b/impl/client/src/main/javascript/web/pentaho/i18n.js
@@ -114,10 +114,10 @@ define([
 
     // Remove basePath from bundle url
     // Taking into account embedded scenarios, where it will be a full URL - "http://host:port/..."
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/pentaho/i18n.js
```diff
@@ -148,4 +146,30 @@ define([
...
+      var scheme = url.protocol;
+      var host = url.host;
+
+      port = defaultSchemePorts[scheme];
+
+      origin = scheme + "//" + host + ":" + port;
+      url = origin + pathname;
+    }
+
+    return new RegExp("^(?:" + url + "|" + origin + "|" + pathname + ")(.*)");
```
dcleao: It might be a good idea to escape `url`, `origin` and `pathname` before concatenation.
See https://github.com/benjamingr/RegExp.escape/blob/master/polyfill.js.

PR #1223, opened 2017-12-06
BACKLOG-19962 - changed dataAttribute to __dataAttribute and fixed te…

RPAraujo:
…sts Added __dataAttribute getColumnProperty to _AbstractTable to be able to remove directa access to __dataAttribute

### DIFF
--- a/impl/client/src/main/doc-js/pentaho/data/ITable.jsdoc
+++ b/impl/client/src/main/doc-js/pentaho/data/ITable.jsdoc
@@ -65,7 +65,7 @@
  */
 
 /**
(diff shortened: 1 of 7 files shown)


impl/client/src/main/doc-js/pentaho/data/ITable.jsdoc
```diff
@@ -64,6 +64,16 @@
  * @return {string} The type name.
  */
 
+/**
+ * Gets a metadata property of a column, given its name and column index.
```
dcleao: Please change to `given the column index and property name.`

PR #1229, opened 2017-12-20
[PRD-5827] - Browser freezes when marking elements on the report toolbar and trying to put data in from keyboard - prevent default handler

AndreyBurikhin:
The fix from https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1221/commits/c67ab3d677f85e821e5e57c8777e6dafcdf61365 doesn't not work for all cases so it should be replaced with a custom event handling.
@tmorgner , @kcruzada, @Aliaksandr-Kastenka, @annapopova1 please review.


impl/client/src/main/javascript/web/dojo/pentaho/common/RowLimitControl.js
```diff
@@ -52,6 +52,9 @@ define(["dojo/_base/declare", "dijit/form/Select", "dijit/form/ValidationTextBox
               this.inherited(arguments);
               on(this.rowLimitRestrictions, 'change', lang.hitch(this, '_onSelect'));
               on(this.rowsNumberInput, 'keydown, focusout', lang.hitch(this, '_onRowLimitSubmit'));
+              on(this.rowsNumberInput, 'mousedown', function(e){ 
+                e.preventDefault();
```
annapopova1: I think will be better to check here disabled or not the input and use `e.preventDefault();` only if the input is disabled.

kolinus:
@tmorgner could you please take a look

PR #1233, opened 2017-12-28
[SP-3894] - Backport of [PIR-1365] Report filter not working correctly

vasilii-komarov:


vasilii-komarov:
@mchen-len-son please review

PR #1234, opened 2018-01-04
[PIR-1365]-Report filter not working correctly

stepanovdg:


stepanovdg:
@afrjorge @dcleao  Could you please take a look?

impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
```diff
@@ -684,6 +684,7 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
         }
         break;
       case pentaho.pda.Column.CONDITION_TYPES.EQUAL:
+      default:
```
dcleao: Is there a new enum value that you wish to support with this code? Then you'd better make that value explicit and leave the default handler untouched? Any other reason for doing this?

stepanovdg: No that relates to regression http://jira.pentaho.com/browse/BACKLOG-20786 
Like there are also different values for operator : contains, end with and etc.
They were missed due to previous change. 
Speaking honestly we could even delete 
 case pentaho.pda.Column.CONDITION_TYPES.EQUAL:
And just return to if/else.

dcleao: Yes please. That will make it clearer that IN is the only operator with a different treatment.
Alternatively, in a switch/case, either include all known values and a `default` for handling unforeseen values or, if some of the known values are passing through the `default`, include these in a comment above the `default`.

stepanovdg:
@dcleao - reworked. Could you please take a look?

PR #1237, opened 2018-01-10
[PIR-1391] Filter Parameter Name can not be saved if condition in (is null; is not null)

Advent51:
- parameter name now hides if (is null | not is null) condition was selected

### DIFF
--- a/impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
+++ b/impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
@@ -255,6 +255,7 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
   },
 
   setFilterType: function(type) {
+    domClass.remove(this.parameterContainer, "filterDialogHidden");
     switch (type) {
       case null:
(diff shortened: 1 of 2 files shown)


Advent51:
@dcleao @pamval @nantunes please review

impl/client/src/main/javascript/web/dojo/pentaho/common/FilterDialog.js
```diff
@@ -523,6 +524,12 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dij
     } else {
       domClass.add(node, "filterDialogHidden");
     }
+    if (this.matchComparator.value === pentaho.pda.Column.CONDITION_TYPES.IS_NULL || this.matchComparator.value === pentaho.pda.Column.CONDITION_TYPES.NOT_NULL) {
+      domClass.add(this.parameterContainer, "filterDialogHidden");
+    }
+    else {
```
dcleao: Just above, you have an if/else formatted in a different way... Please be more careful in the future, so that we all don't spend time on these things :-(

Advent51:
@dcleao please review it again :-)

PR #1243, opened 2018-01-29
[PRD-5827] - Browser freezes when marking elements on the report toolbar and trying to put data in  from keyboard.

Aliaksandr-Kastenka:
Additional pr to https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1229
Fix for the case when user select label and input data
Goes with https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/685


Aliaksandr-Kastenka:
@mchen-len-son 
Please review.

PR #1246, opened 2018-02-07
 [BACKLOG-21192] VizAPI 3.0: Visual Roles working modes and mapping strategies

dcleao:

There's a unit test related with the ccc/visual which is failing, but will only be fixed in a later story.


impl/client/src/main/javascript/web/pentaho/data/Attribute.js
```diff
@@ -116,27 +116,50 @@ define([
        */
       this.type = type;
 
-      var isDiscrete = type !== "number" && type !== "date";
-      if(!isDiscrete && spec.isDiscrete != null)
-        isDiscrete = !!spec.isDiscrete;
+      var isContinuous;
+      if(type === "string" || type === "boolean") {
```
nantunes: This default takes precedence even if `spec.isContinuous` is defined. Is that intended? (I understand it makes sense to force as categorical something continuous but not the other way around... my question is if we should "auto-correct" the spec choice at this level)

impl/client/src/main/javascript/web/pentaho/data/Attribute.js
```diff
@@ -116,27 +116,50 @@ define([
...
       /**
-       * Indicates if the attribute is considered discrete.
+       * Indicates if the attribute is considered **continuous**, or,
+       * on the other hand, **categorical**.
        *
        * When the attribute's type is one of
        * {@link pentaho.data.AtomicTypeName.STRING} or
        * {@link pentaho.data.AtomicTypeName.BOOLEAN}
-       * this property is always `true`.
+       * this property is **ignored** and the attribute is necessarily **categorical**.
```
nantunes: Ok, next time I will read the docs before the implementation. ;-)

My question above is answered: it was intentional. I'll just leave the comment to discuss the choice.

impl/client/src/main/javascript/web/pentaho/lang/DomNode.js
```diff
@@ -0,0 +1,168 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+define([
+  "module",
+  "pentaho/lang/Base",
+  "pentaho/util/object"
+], function(module, Base, O) {
+
+  var DomNode = Base.extend(module.id, /** @lends pentaho.lang.DomNode# */{
```
nantunes: What is this? Wasn't the Scene just an util?

impl/client/src/main/doc-js/pentaho/visual/role/strategies/IMapper.jsdoc
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @name pentaho.visual.role.strategies.IMapper#mode
+ * @type {!pentaho.visual.role.Mode}
+ * @readOnly
+ */
+
+/**
+ * Gets a label that describes the visual role as it is mapped.
+ *
+ * @name pentaho.visual.role.strategies.IMapper#label
```
nantunes: I still see this as a function and not as a property of the mapper... But maybe I'm wrong. Is a mapper instance immutable? Will any change in the VR fields or the input data necessarily create a totally new mapper?

impl/client/src/main/doc-js/pentaho/type/spec/ITypeDescriptor.jsdoc
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The `spec.ITypeDescriptor` interface describes a type descriptor in _object_ form.
```
nantunes: Why was the TypeDescriptor needed? I remember vaguely... It had to do with the data table types vs. the type api types?

impl/client/src/main/javascript/web/pentaho/data/Attribute.js
dcleao: String is inherently categorical. The only way it could be seen as continuous is if some conversion could be performed on it. Because the data table offers no means for such conversion, then it is necessarily categorical.
With, boolean, we could also see it in a continuous numeric scale, if we would map `false` to `0` and `true` to `1`, for example. Still, a conversion is being assumed.
I don't see this as an automatic correction, but more like the `spec.isContinuous` option only applies (ignored otherwise) to types which _can be_ continuous, and for some types, we know they cannot be.

impl/client/src/main/javascript/web/pentaho/lang/DomNode.js
dcleao: This is the `DomNode` class. `Scene` inherits from `DomNode`. Scene is optional and can be seen as "just an utility". Scenes are inherently hierarchical, so DOM-like functionality is part of its functionality. Defining a `DomNode` base class helps to keep DOM stuff separate from the `Scene`'s functionality which is purely visual related. Additionally, a `DomNode` class is useful in its own, and has a language-level character.

impl/client/src/main/doc-js/pentaho/visual/role/strategies/IMapper.jsdoc
dcleao: I didn't review this name or the type of member since we talked about it, in the other day.
The arguments from which a mapper is created are immutable. They do have internal mutable state to be able to satisfy the `invertValue` method. A change to the data or the VR mappings will cause mappers to be invalidated and new mapper instances will be obtained, the next time the `Mapping#mapper` property is read.

So, the label would be immutable for each mapper instance, but `this.model.myRole.mapper.label` can still change over time.

Maybe `expressionLabel`? In the sense that the type of supported mappings corresponds to a combination of the attributes' values that can be written as an expression?

impl/client/src/main/doc-js/pentaho/type/spec/ITypeDescriptor.jsdoc
dcleao: `TypeDescriptor` is `Simple` type which wraps a `pentaho.type.Type` instance. It allows a property of a complex type to hold a type. The use case is that of `pentaho.visual.role.Mode#dataType`. Then `pentaho.visual.role.Property.Type#modes` — a type attribute — holds a list of `Mode` instances. Up until now, we didn't have a use case like this; types always referenced other types directly and not through a complex value.

impl/client/src/main/javascript/web/pentaho/visual/scene/Base.js
```diff
@@ -102,11 +120,30 @@ define([
...
+     *
+     * @param {!pentaho.visual.scene.Base} child - The child scene.
+     * @return {number} The index at which the given child scene was inserted.
      * @private
      */
-    __setAsRoot: function() {
-      O.setConst(this.vars, "$root", this);
+    __appendChildCore: function(child) {
+
+      var children = O.getOwn(this, "children") || (this.children = []);
```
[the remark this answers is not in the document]
dcleao: `O.getOwn` only returns local - i.e. _own_ - properties, not inherited ones. Thus, until the array is created locally, the function will return `undefined`. The prototype's `Object.freeze([])` exists only to avoid allocating arrays for each leaf scene, while still providing the convenience to "consumers" to not have to test for `null` children before iteration, in generic code.

PR #1250, opened 2018-03-06
[BACKLOG-16958] Removes node_modules dir during clean

axelguiloff:
@pentaho/x-wing Please review

### DIFF
--- a/impl/client/pom.xml
+++ b/impl/client/pom.xml
@@ -19,6 +19,7 @@
   <properties>
     <js.project.list>dojo,dijit,jquery,pentaho-cdf-js,dojo-release,angular,angular-mocks,angular-resource,angular-route,angular-animate</js.project.list>
     <build.javascriptReportDirectory>target/js-reports</build.javascriptReportDirectory>
+    <maven-clean-plugin.version>2.5</maven-clean-plugin.version>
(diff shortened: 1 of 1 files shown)


tmcsantos:
Couldn't this be done in the parent poms?
I think there are more projects with the same issue.

graimundo:
@tmcsantos I agree that this should be done in the parent poms. This is most likely affecting more projects other than common-ui. What's the profile that is being executed that installs node modules? Most likely the clean plugin should be put in that profile as well.

@diogofscmariano any inputs on this topic?

tmcsantos:
I believe the profile active profile would be this one. https://github.com/pentaho/maven-parent-poms/blob/master/pom.xml#L520

diogofscmariano:
@graimundo no, just the fact that if we are on a clean environment, deleting that will cause all the node/npm environment to be reinstalled. IMO, the right approach would be to put this inside the target folder, but we already had problems/discussions in the past with that

graimundo:
@diogofscmariano I agree with you. Now that we have more experience and use cases of projects that are using node modules It's probably time to revisit the subject of where node modules should be put and when they should be cleaned. I would put this PR on hold until that discussion takes place Does @pentaho/x-wing agree?

axelguiloff:
Closing this PR pending further discussion

PR #1255, opened 2018-03-16
[BACKLOG-22104] DEV: avoid select filter full redraw when feeding data

afrjorge:
- add widget parameter to allow it's builder to react to parameter value changes
  - stops executing updates when the CDF component has 'isManaged' set to 'false'


afrjorge:

PR #1258, opened 2018-03-19
[BACKLOG-21717] Common-ui webpackage assembly

nantunes:

### DIFF
--- a/assemblies/webpackage/pom.xml
+++ b/assemblies/webpackage/pom.xml
@@ -10,8 +10,7 @@
   </parent>
 
   <artifactId>common-ui-webpackage</artifactId>
-  <version>8.1.0.0-SNAPSHOT</version>
-
+         
   <packaging>bundle</packaging>
 
   <properties>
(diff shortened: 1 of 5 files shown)


assemblies/webpackage/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+
+        <executions>
+          <execution>
+            <id>unpack-karaf-pkg</id>
```
davidmsantos90: This id could be more suggestive, with like 'unpack-common-ui-pkg'

assemblies/webpackage/src/main/resources-filtered/common-ui/i18nMock.js
```diff
@@ -0,0 +1,69 @@
+/*!
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
```
davidmsantos90: the copyright header is outdated

PR #1259, opened 2018-03-21
[PIR-1397] Looping queries on Filter dialog due to lack of error handling

Advent51:
- looping was avoided
- error message was added


Advent51:

PR #1265, opened 2018-04-03
[BACKLOG-21825] - Added unit tests for modelAdapter._convertFilterToI…

graimundo:
…nternal

- Fixed bugs for edge cases and made conversion less lenient for filters not explicitly considered.

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/visual/base/modelAdapter.js
+++ b/impl/client/src/main/javascript/web/pentaho/visual/base/modelAdapter.js
@@ -93,10 +93,11 @@ define([
     "./model",
     "../role/externalProperty",
     "pentaho/data/filter/true",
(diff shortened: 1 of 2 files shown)


graimundo:

impl/client/src/main/javascript/web/pentaho/visual/base/modelAdapter.js
```diff
@@ -589,77 +604,102 @@ define([
...
+                  __context
+                );
+              }, this));
             }
-          });
 
-          if(equalsMap === null) {
-            return filter;
-          }
+            return new AndFilter({operands: operands});
```
nantunes: What happens if `equalsMap` is empty? Is an empty `and` equivalent to `true`, like in `isEqual` case bellow?

PR #1267, opened 2018-04-04
[BACKLOG - 21825] Test view is dirty on model change

RPAraujo:

### DIFF
--- a/impl/client/src/test/javascript/pentaho/visual/base/view.spec.js
+++ b/impl/client/src/test/javascript/pentaho/visual/base/view.spec.js
@@ -1011,6 +1011,24 @@ define([
 
         expect(view.isDirty).toBe(true);
       });
+
+      it("should become dirty on model change", function() {
+
+        view.__dirtyPropGroups.clear(); // view is clean
+
+        view.isAutoUpdate = false;
+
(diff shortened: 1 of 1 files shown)


impl/client/src/test/javascript/pentaho/visual/base/view.spec.js
```diff
@@ -1011,6 +1011,24 @@ define([
...
       });
+
+      it("should become dirty on model change", function() {
+
+        view.__dirtyPropGroups.clear(); // view is clean
+
+        view.isAutoUpdate = false;
+
+        var didChangeHandler = jasmine.createSpy().and.callFake(function(){
+          expect(view.isDirty).toBe(true);
```
nantunes: Why is the `didChangeHandler ` needed? Couldn't this be tested bellow, after changing `model.selectionFilter`?

dcleao: The idea is making sure that the view knows it is dirty at that time — even before it's own did:change handler, that notifies it of the model having changed, is called.
Afterwards, it is surely dirty, and other tests already test(ed) that.

nantunes: 👍

PR #1274, opened 2018-04-17
[PIR-1400] UI - Filter dialog box issue. Value field moves to the next row

Advent51:
- UI was updated according to the mock ups provided.
- there is an issue with "text-overflow: ellipsis" in IE 11 but I have no idea how to fix it. (ellipsis is not rendered)


Advent51:

PR #1277, opened 2018-04-19
[PIR-1400] UI - Filter dialog box issue. Value field moves to the next row

Advent51:
- additional UI changes were introduced:
- close dialog button (X) was removed
- fixed the layout issue with Crystal theme
- little code cleanup


Advent51:

PR #1296, opened 2018-05-21
[PRD-5963] - Date Picker showing One Day Prior for Default Value of =…

cravobranco:
…TODAY and =YESTERDAY when client is in Different Timezone

   * Added client side support to handle reports which date is calculated based on client timezone
   * Refactored the date parser based on timezone type value, which can be one of the following:
    1) server - timezone already handled by the server
    2) client - timezone handled in the client side (see pentaho-platform-plugin-common-ui - formatting.js)
    3) utc - timezone forced to UTC
    4) customized timezone - timezone forced to customized timezone


cravobranco:
@pentaho-lmartins , @pamval , @mbatchelor , @mchen-len-son , @tmorgner  please review and merge if it's all ok

Notes: This should be reviewed together with:
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/627

Main changes:
- Added client side support to handle reports which date is calculated based on client timezone
- Refactored the date parser based on timezone type value, which can be one of the following:
1) server - timezone already handled by the server
2) client - timezone handled in the client side (see pentaho-platform-plugin-common-ui - formatting.js)
3) utc - timezone forced to UTC
4) customized timezone - timezone forced to customized timezone

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -192,6 +192,30 @@ define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/
...
           if(timezone == 'client') {
+
+              if (value.length == 28) {
+
+                  var currentDate = new Date();
+                  var offset = -currentDate.getTimezoneOffset();
+
+                  // calculation is done in minutes because it is the smaller unit on a timezone definition
+                  // e.g.: "2018-05-20T01:12:23.456-0430" where -04h30m is the timezone setting
+                  var offsetServerMinutes = value.substring(23, 26) * 60;
```
dcleao: The piece you extract is something like `0430`, which represents `4*60 + 30 = 270` minutes. The code you wrote would interpret this as 430 minutes.
Also, you seem to be ignoring offset's +/- sign (and why the `-` on the offset variable above?).

Also, it is "not considered good practice" to let JavaScript implicitly convert from string to number — it's usually best to write something more explicit, such as `Number(text)` or an explicit `+` sign, as in `(+text)`.

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -192,6 +192,30 @@ define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/
           }
 
           if(timezone == 'client') {
+
+              if (value.length == 28) {
+
+                  var currentDate = new Date();
+                  var offset = -currentDate.getTimezoneOffset();
```
dcleao: Perhaps calling this `offsetClientMinutes` would be clearer.

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -192,6 +192,30 @@ define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/
           }
 
           if(timezone == 'client') {
+
```
dcleao: So are you handling case 4?
> 4. customized timezone - timezone forced to customized timezone

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -192,6 +192,30 @@ define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/
...
+                  var diffOffset = offset - offsetServerMinutes;
+
+                  if(diffOffset != 0){
+                      this._initDateFormatters();
+
+                      var currentMinutes = ( currentDate.getHours() * 60 ) + currentDate.getMinutes();
+
+                      var diffMinutes = currentMinutes - diffOffset;
+                      var localDate = this.parseDateWithoutTimezoneInfo(value);
+                      var dayVariance = (diffMinutes < 0) ? 1 : (diffMinutes >= 1440) ? -1 : 0;
```
dcleao: Magic constants?

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -192,6 +192,30 @@ define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/
...
+
+                  // calculation is done in minutes because it is the smaller unit on a timezone definition
+                  // e.g.: "2018-05-20T01:12:23.456-0430" where -04h30m is the timezone setting
+                  var offsetServerMinutes = value.substring(23, 26) * 60;
+                  var diffOffset = offset - offsetServerMinutes;
+
+                  if(diffOffset != 0){
+                      this._initDateFormatters();
+
+                      var currentMinutes = ( currentDate.getHours() * 60 ) + currentDate.getMinutes();
```
dcleao: I cannot see why the number of minutes into the day is relevant...

impl/client/src/main/javascript/web/util/formatting.js
cravobranco: No, in this code snippet I am handling case 2 (i.e. when the display is based on client timezone).
Regarding case 4, it's only handled in server side and it was implemented in this same delivery but in the server's repository, please check: https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/627

impl/client/src/main/javascript/web/util/formatting.js
cravobranco: Agree, please see the new updated commit.

impl/client/src/main/javascript/web/util/formatting.js
cravobranco: Agree, please see the new updated commit.
I added a global variable const : totalDayMinutes = 1440

impl/client/src/main/javascript/web/util/formatting.js
cravobranco: Actually in the example above I am using `-04` substring (and then transforming it to `-04*60`), which is also wrong because the intention is to use also the minutes part.

Agree with the `Number(text)` suggestion, so I will use the following statement instead:
`var offsetServerMinutes = Number( (value.substring(23, 26) * 60) + value.substring(26, value.length()) );`

Regarding the `-` signal, the reason is that javascript's method getTimezoneOffset() returns the value as negative in accordance to the timezone.
Example: If your time zone is GMT+2, -120 will be returned (more info on https://www.w3schools.com/jsref/jsref_gettimezoneoffset.asp)

impl/client/src/main/javascript/web/util/formatting.js
cravobranco: The number of minutes are used in order to avoid to loose precision, because timezone's granularity is in minutes (e.g. -04:30)

cravobranco:
@dcleao thanks for your comments.
I committed just now your suggestions and squashed it.
Please see my explanations regarding the raised doubts.
Please review also the new changes and merge if it's all ok.

impl/client/src/main/javascript/web/util/formatting.js
dcleao: You're right about the substring capturing `-04` and not `0430`...
I also did not know about the `getTimezoneOffset` returning the negative of what I expected.
About the use of `Number`, note that you're still relying on the implicit conversion when doing `(value.substring(23, 26) * 60)`.

impl/client/src/main/javascript/web/util/formatting.js
```diff
@@ -20,6 +20,9 @@
  */
 define("common-ui/util/formatting", ['common-ui/util/timeutil', 'common-ui/util/TextFormatter'], function(ReportTimeUtil, TextFormatter) {
   return {
+
+    const : totalDayMinutes = 1440,
```
dcleao: Please create this as a var, just before the return statement.
The way this is, you're defining a property named "const" in the exported object and defining a global variable called `totalDayMinutes`.

```js
var totalDayMinutes = 1440;

return {
  // ...
};
```

PR #1306, opened 2018-07-10
[BACKLOG-23472] Adds `pentaho/theme` AMD loader plugin.

dcleao:

### DIFF
--- a/impl/client/src/main/javascript/web/pentaho/theme/main.js
+++ b/impl/client/src/main/javascript/web/pentaho/theme/main.js
@@ -80,7 +80,7 @@ define([
           : normalizedName;
 
         configService.selectAsync("pentaho/theme!" + targetModuleId)
-          .then(function(themeConfig) { return loadTheme(themeConfig, targetModuleId); })
+          .then(loadTheme)
           .then(function() { onLoad(); }, onLoad.error);
       }
     },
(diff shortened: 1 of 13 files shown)


impl/client/src/main/javascript/web/pentaho/theme/main.js
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * To register a theme module, configure the module composed of the plugin and its argument.
+   * For example, to configure the theme of the `my/View` module to be the `my/theme/view` module,
+   * specify the following AMD/RequireJS configuration:
+   *
+   * ```js
+   * {
+   *   "config": {
+   *     "pentaho/theme!my/View": {
+   *       main: "my/theme/view"
```
nantunes: All these moduleIDs must be versioned by a requirejs manager plugin.

dcleao:
@davidmsantos90 As for the `package-lock.json`, I really intended adding it.
As per the opinion of who invented the file, it should be checked-in to source control: https://docs.npmjs.com/files/package-lock.json.

impl/client/src/main/javascript/web/pentaho/theme/main.js
dcleao: We've talked in the meantime. The conclusion is that the `config` keys are already being versioned. As for the the values, a custom plugin needs to exist that knows the structure of the actual configuration. We'd move on without such a plugin for now, as users can still use configuration file rules to achieve the same effect, and in that context a means will exist to version these ids (by using `module.resolveId()`).

PR #1310, opened 2018-09-03
[PRD-6019] - After deleting the input field on the parameter UI, the …

ricardosilva88:
…mandatory parameters validation no longer works and no warning message occurs

### DIFF
--- a/impl/client/src/main/javascript/web/prompting/PromptPanel.js
+++ b/impl/client/src/main/javascript/web/prompting/PromptPanel.js
@@ -1141,6 +1141,7 @@ function(Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashb
             for(var i in existingErrors) {
               _removeChildComponent.call(this, panel, errComp);
             }
+            errComp.clearAndRemove = true;
             toRemove.push(errComp);
           }
         }
@@ -1159,8 +1160,9 @@ function(Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashb
               var errIndex = panel.components.length - 1;
               var errorComponent = _createWidgetForErrorLabel.call(this, param, error);
               this.dashboard.addComponent(errorComponent);
-              this.dashboard.updateComponent(errorComponent);
               panel.components.splice(errIndex, 0, errorComponent);
+              panel.addErrorLabel(errorComponent);
+              this.dashboard.updateComponent(errorComponent);
             }
           }
         }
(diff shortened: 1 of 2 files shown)


impl/client/src/main/javascript/web/prompting/PromptPanel.js
```diff
@@ -1161,6 +1162,7 @@ function(Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashb
               this.dashboard.addComponent(errorComponent);
               this.dashboard.updateComponent(errorComponent);
               panel.components.splice(errIndex, 0, errorComponent);
+              panel.addErrorLabel(errorComponent, error);
```
ricardosilva88: As discussed with @dcleao, added this function to add error message in a div label to the panel, avoiding a full reload of prompt panel

impl/client/src/main/javascript/web/prompting/components/ParameterPanelComponent.js
```diff
@@ -26,6 +26,23 @@ define("common-ui/prompting/components/ParameterPanelComponent", ['./PanelCompon
...
+      $htmlObject.addClass('error');
+    },
+
+    addErrorLabel: function(errComponent, errMessage){
+      var $htmlObject = this.placeholder();
+      if(!errComponent || !errMessage)
+        return;
+
+      var errorLabelHtml = '<div id="' + errComponent.htmlObject + '" class="parameter-label">';
+      errorLabelHtml += errMessage + '</div>';
```
dcleao: I don't think you should be injecting the error message like this. Is it sanitized? How is it done in the full-render mode?

ricardosilva88: In full render mode is done like is described in the following example js https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/impl/client/src/main/javascript/web/prompting/components/CompositeComponent.js, you don't have a function receiving directly an error message sting. But as we discussed previously, this would be a very "special" case of creating and injecting an label on a panel component in opposite to all CDF logic where all components were static, that why I opted by this solution. In your opinion what would be the ideal solution?

dcleao: You should be creating an empty div for the provided errComponent. It is the component that, when updated, should fill-in the div with the error message. Don't know how the specific component receives the error message.
You need to update the component afterwards.

ricardosilva88:
PR update with changes that we discussed

PR #1311, opened 2018-09-04
[PIR-1416] When hovering a warning message on PIR a message with code…

LeonardoCoelho71950:
… is displayed

Previously, it was used a title attribute, to define the title. But, the title attribute specifies extra information, shown as a tooltip text, when the mouse moves over the element.

@pamval @duarteteixeira 
@ricardosilva88 @pentaho-lmartins


pentaho-lmartins:
@pentaho/millenniumfalcon and @dcleao, can you take a look?

PR #1322, opened 2018-10-06
[BACKLOG-24643] Fix regression caused in DET.

dcleao:


dcleao:
Merging this now, to unblock today's QAT.

PR #1378, opened 2019-02-05
[PPP-4223] - Use of vulnerable component spring-core before 4.3.14, s…

ricardosilva88:
**DO NOT MERGE WITHOUT THE REST OF PRs:**
[pentaho/maven-parent-poms#95 (comment)](https://github.com/pentaho/maven-parent-poms/pull/95#issue-250486125)

### DIFF
--- a/impl/server/pom.xml
+++ b/impl/server/pom.xml
@@ -13,7 +13,6 @@
   <description>Java portion of the Common UI codebase</description>
   <url>http://www.pentaho.com</url>
   <properties>
-    <spring.version>4.3.2.RELEASE</spring.version>
     <commons-vfs2.version>2.2</commons-vfs2.version>
     <simple-jndi.version>0.11.1</simple-jndi.version>
     <jmock-junit4.version>2.5.1</jmock-junit4.version>
@@ -97,11 +96,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+
+    <!-- Needed for compilation -->
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-core</artifactId>
-      <version>${spring.version}</version>
-      <scope>provided</scope>
       <exclusions>
         <exclusion>
           <artifactId>*</artifactId>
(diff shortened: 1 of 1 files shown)


impl/server/pom.xml
```diff
@@ -97,23 +96,23 @@
         </exclusion>
       </exclusions>
     </dependency>
+
+    <!-- Needed for compilation -->
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-core</artifactId>
-      <version>${spring.version}</version>
-      <scope>provided</scope>
```
graimundo: Doesn't removing the provided scope have an impact in the plugin assembly? Isn't spring included in the plugin lib?

impl/server/pom.xml
```diff
@@ -97,23 +96,23 @@
...
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- -->
+
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-beans</artifactId>
-      <version>${spring.version}</version>
-      <scope>provided</scope>
```
graimundo: Doesn't removing the provided scope have an impact in the plugin assembly? Isn't spring included in the plugin lib?

impl/server/pom.xml
ricardosilva88: Not sure why a remove it, will test it, but is very likely to maintain provided here

impl/server/pom.xml
ricardosilva88: Not sure why a remove it, will test it, but is very likely to maintain provided here

ricardosilva88:
PR updated

impl/server/pom.xml
ricardosilva88: At first I kept provided scope because it was already there, and then I believe a did some testing on it and opted to keep provided. Since the outcome of this project will be a plugin deployed on the server, is most likely that spring-core dependency is being provided by server context. Do you want me to invest some more testing on it?

impl/server/pom.xml
ricardosilva88: At first I kept provided scope because it was already there, and then I believe a did some testing on it and opted to keep provided. Since the outcome of this project will be a plugin deployed on the server, is most likely that spring-beans dependency is being provided by server context. Do you want me to invest some more testing on it?

impl/server/pom.xml
ricardosilva88: Spring jars aren't going to /lib folder of plugin

impl/server/pom.xml
ricardosilva88: Spring jars aren't going to /lib folder of plugin

PR #1382, opened 2019-02-12
[BISERVER-13901] When auto-submit is off, the report parameter/prompt panel flashes/gl…

LeonardoCoelho71950:
…ass pane appears when parameter is TEXT field.

Depends on https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/682

@pentaho/tatooine @pentaho-lmartins 
@dcleao

### DIFF
--- a/impl/client/src/main/javascript/web/prompting/PromptPanel.js
+++ b/impl/client/src/main/javascript/web/prompting/PromptPanel.js
@@ -1405,7 +1405,7 @@ function(Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashb
 
         // Must submit, independently of auto-submit value.
         // All parameters are initialized, fire the submit
-        fireSubmit = !noAutoAutoSubmit;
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/prompting/PromptPanel.js
```diff
@@ -1405,7 +1405,7 @@ function(Base, Logger, DojoNumber, i18n, Utils, GUIDHelper, WidgetBuilder, Dashb
 
         // Must submit, independently of auto-submit value.
         // All parameters are initialized, fire the submit
-        fireSubmit = !noAutoAutoSubmit;
+        fireSubmit = (noAutoAutoSubmit != undefined) ? !noAutoAutoSubmit : false;
```
dcleao: Maybe it's clearer this way? Also, when comparing with `==` (or `!=`) it's best (and equivalent) to compare with `null`.
```suggestion
        fireSubmit = (noAutoAutoSubmit != null) && !noAutoAutoSubmit;
```

PR #1406, opened 2019-04-04
[BACKLOG-28349] Use of vulnerable component timespan 2.3.0 CVE-2017-1…

dcleao:
…6115

* Updating karma to the latest version removed that and other vulnerabilities.
* Also, removed the unused karma plugins: `karma-html-reporter` and `karma-jasmine-html-reporter`.


dcleao:
@lucboudreau, the updated components in this PR are only used at build and test-time. They are not shipped with the product (and, thus, should not generate a component process).

PR #1426, opened 2019-05-28
[PIR-1442] Activating auto-refresh after adding a prompt in an Interactive Report causes report to become non-responsive

hugo-jose:
Bug fix: Activating auto-refresh after adding a prompt in an Interactive Report causes report to become non-responsive


hugo-jose:

hugo-jose:
@ricardosilva88 @RPAraujo @ssamora Please review

PR #1437, opened 2019-06-04
[PIR-1442] Activating auto-refresh after adding a prompt in an Interactive Report causes report to become non-responsive

hugo-jose:


hugo-jose:
@pamval @kcruzada please merge

PR #1631, opened 2023-03-03
[BACKLOG-36893] Changes for WCAG Focus Order (A)

AAravind27:


AAravind27:
@pentaho/wcag Please review

impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/Dialog.js
```diff
@@ -0,0 +1,674 @@
...
+            focus = pd.focus;
+            if(pd.dialog && (!focus || !dom.isDescendant(focus, pd.dialog.domNode))){
+              pd.dialog._getFocusItems(pd.dialog.domNode);
+              focus = pd.dialog._firstFocusItem;
+					  }
+					}
+
+					if(focus){
+						// Refocus the button that spawned the Dialog.   This will fail in corner cases including
+						// page unload on IE, because the dijit/form/Button that launched the Dialog may get destroyed
```
dcleao, line 603: I believe formatting is completely off, compared to the original. Could you have auto-formatted the file? In my local version, I see a 2 space indented file. This change will make it harder (now and) in the future to know exactly what changed. And, in such a big file, it's important.

So, please, restore the original formatting and add your modifications following the original formatting.

Also, please, to the same end of making it easier to discover changes, add a comment before changed lines including the word "Pentaho": For example:
```js
// Pentaho - adding support for a fixed dialog opener.
```

dcleao: Also, some of the files within the overrides folder have a header such as:

```js
/* Overridden due to PRD-5440
		see lines: 200-205
*/
```

I'd use instead something like:
```js
/* Overridden due to PRD-5440
		see comments starting with // Pentaho
*/
```

impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/Dialog.js
```diff
@@ -0,0 +1,674 @@
...
+						dialog.dialogOpener = null;
+					}else{
+            // If we are returning control to a previous dialog but for some reason
+            // that dialog didn't have a focused field, set focus to first focusable item.
+            // This situation could happen if two dialogs appeared at nearly the same time,
+            // since a dialog doesn't set it's focus until the fade-in is finished.
+            focus = pd.focus;
+            if(pd.dialog && (!focus || !dom.isDescendant(focus, pd.dialog.domNode))){
+              pd.dialog._getFocusItems(pd.dialog.domNode);
+              focus = pd.dialog._firstFocusItem;
```
dcleao, line 597: Where is the code where a menu item's parent would be followed until a stable anchor element to place focus on were found?

dcleao (changes requested):

impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/Dialog.js
AAravind27: Hi, Duarte. Turns out that code was not needed when setDialogOpener was introduced

dcleao: You can get away with dialog opener in every case, as long as you handle all instances explicitly — that was a given when we talked first.

I believe the goal would be to achieve a default behavior that would cover most cases, while having the dialog opener to cover the ones which the default behavior could not.

Is there any idea for how many cases are there that would benefit from this default behavior? Every dialog being opened from an Analyzer context menu?

impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/Dialog.js
dcleao: @AAravind27 I don't think that in your last commit you have restored the original formatting as I asked for:

> So, please, restore the original formatting and add your modifications following the original formatting.

dcleao: I checked the original file and it seems that, unlike our own codebase, it uses tabs instead of spaces, and that's why I thought it was looking strange.
Unfortunately, this also means that your new additions/changes would preferably use tabs to match.

dcleao (changes requested):
All good, but found out about another detail that needs to be handled when overriding Dojo JS files. Please see this PR for a complete example: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1633.

dcleao (approved):

dcleao:
When merging this PR, please **squash** the commits!

PR #1632, opened 2023-03-06
[BACKLOG-37162] Fixing Keyboard Navigation (A) for color palette

hv-vamsi:
@pentaho/wcag, Please review.

### DIFF
--- a/impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/_PaletteMixin.js
+++ b/impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/_PaletteMixin.js
@@ -1,0 +1,357 @@
+/* Overridden due to [BACKLOG-36890]
+		see lines: 170-185
+*/
+
+define([
+	"dojo/_base/declare", // declare
+	"dojo/dom-attr", // domAttr.set
+	"dojo/dom-class", // domClass.add domClass.remove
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/_PaletteMixin.js
```diff
@@ -0,0 +1,357 @@
...
+		},
+
+		postCreate: function(){
+			this.inherited(arguments);
+
+			// Set initial navigable node.
+			var cell = this._getValueCell();
+			if (cell) {
+				this._setCurrent(cell.node);
+			}
```
dcleao, line 171: Should likely default to the first cell, `this._cells[0]`, like in the original behavior, when `value` is not defined or cannot be matched with the existing palette.
```js
var cell = this._getValueCell() || this._cells[0];
```

dcleao (changes requested):

impl/client/src/main/javascript/web/dojo/pentaho/common/overrides/dijit/_PaletteMixin.js
hv-vamsi: addressed

dcleao (approved):

addagudi (approved):

phalguni-a (approved):

PR #1634, opened 2023-03-09
[BACKLOG-36891] Fixed NPE when a11y utilities are given null elements

dcleao:
- Opting for ease of use of the utility by tolerating null elements being passed in

The changes in this story caused a regression (NPE at startup, impeding it) when Analyzer is used with options (e.g. http://localhost:8080/pentaho/api/repos/%3Apublic%3ASteel%20Wheels%3ACountry%20Performance%20(heat%20grid).xanalyzer/editor?disableFilterPanel=true). This also affects PDI/DET.

@pentaho/wcag, please, review.

### DIFF
--- a/impl/client/src/main/javascript/web/util/_a11y.js
+++ b/impl/client/src/main/javascript/web/util/_a11y.js
@@ -42,13 +42,10 @@ define(function () {
     elem.addEventListener('keydown', actionButtonKeyDownHandler);
     elem.addEventListener('keyup', actionButtonKeyUpHandler);
 
-    dispose.remove = dispose;
-    return dispose;
-
-    function dispose() {
+    return createDisposable(function dispose() {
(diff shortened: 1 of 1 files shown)


impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -84,12 +81,34 @@ define(function () {
...
+    });
+  }
+
   return {
     keyCodes: keyCodes,
+
     makeAccessibleActionButton: function (elem) {
+      if (elem == null) {
+        return createNullDisposable();
+      }
```
dcleao, line 102: Ease of use comes with also having to return a disposable, even though a dummy one.

singletonc (approved):

dcleao:
Merging to unblock work. Please, still, make your review. I'll issue a follow-up PR with any corrections.

PR #1639, opened 2023-03-30
[BACKLOG-37351] Change modules to be RequireJS first and to export to global variables later

dcleao:
@pentaho/wcag, please, review.

[BACKLOG-37351](https://hv-eng.atlassian.net/browse/BACKLOG-37351)

The problem occurred because, in the unit test environment, jQuery is not loaded before the `common-ui/util/_focus` and `common-ui/util/_dialog` scripts. In this environment, these are loaded like any other RequireJS modules, however, the scripts are not regular RequireJS modules and do not declare the dependency on jQuery. In the server, they are loaded instead as static resources, always after jQuery is loaded, by `webcontext.js` (in the `global` context), which is one of the first things to be included by any HTML page (Mantle.jsp, Analyzer, PDD, ...).

- Using the same dual Global & RequireJS module pattern as `common-ui/util/URLEncoder`.
- _In principle_, the global variables get defined before any use.
- The RequireJS modules start "loading" when the scripts are loaded by webcontext.js, and the corresponding global variables will be available only when the RequireJS modules load.
- The first known use of the corresponding global variables is by `MantleLoginEntryPoint`, which eagerly creates the `MantleLoginDialog`, which then uses `pho.util._dialog`. I believe that this always occurs many event loop turns after the dependencies are made globally available (and if we find an issue here, we could easily instead lazily create the login dialog).
- There appears to be no easy way for a GWT module to depend on a RequireJS module and to incorporate that within the GWT module loading itself (e.g. loading of essential i18n resources is typically done on `EntryPoint#onModuleLoad()` and it has to be the application to specially handle the additional wait). Ideally, we'd be able to load the modules `common-ui/util/_focus` and `common-ui/util/_dialog` within the scope of `WidgetModule` (an `EntryPoint`), where the classes `ElementUtils` and `DialogBox` which make use of the latter live. Because that does not seem to be possible, in the past, we've resorted to the use of these dual module format scripts and on the empirically proven availability of the resources. IMO, a good solution until proven otherwise (or given enough time/human resources to solve the issue at a deeper level, somehow).

Related PR for `wcag` branch: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1638

[BACKLOG-37351]: https://hv-eng.atlassian.net/browse/BACKLOG-37351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/impl/client/src/main/javascript/web/util/_dialog.js
+++ b/impl/client/src/main/javascript/web/util/_dialog.js
@@ -20,23 +20,19 @@
  * Portions of this file are based on jQuery UI, v1.13.2
  */
 
-var pho = pho || {};
-if (pho.util == null) {
-  pho.util = {};
-}
-
-(function() {
+/* Need module name as this file is loaded first as a static resource by common-ui. */
+define("common-ui/util/_dialog", [
+  "./_focus",
+  "../jquery-clean"
+], function(focusUtil, $) {
   "use strict";
-
-  /** @type {jQuery} */
-  var jQueryLocal = pho.util._focus.jQuery;
 
   var KeyCodes = {
     tab: 9
   };
   var Selectors = {
     autoFocus: "[autofocus]",
-    tabbable: pho.util._focus.Selectors.tabbable
+    tabbable: focusUtil.Selectors.tabbable
   };
   var RestoreFocusModes = {
     off: 1,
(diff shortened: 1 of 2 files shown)


buildguy:
### :white_check_mark: Build finished in 5m 19s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
:ok_hand: All tests passed!
 
Tests run: **73**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/common-ui/job/PR-1639/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

PR #1642, opened 2023-04-11
[BACKLOG-37343] Listbox changes using JS. Removed the changes made with select and option.

hv-vamsi:
@pentaho/wcag Please review

### DIFF
--- a/impl/client/src/main/javascript/web/util/_a11y.js
+++ b/impl/client/src/main/javascript/web/util/_a11y.js
@@ -18,6 +18,7 @@
 define(["./_focus"], function(focusUtil) {
 
   var keyCodes = {
+    A: 65,
     backSpace: 8,
     tab: 9,
     enter: 13,
@@ -34,6 +35,385 @@ define(["./_focus"], function(focusUtil) {
     arrowDown: 40,
     delete: 46
   };
+
+  var keys = {
+    a: "a",
+    A: "A",
+    Backspace: "Backspace",
+    Tab: "Tab",
+    Enter: "Enter",
+    Shift: "Shift",
+    Escape: "Escape",
+    Space: " ",
+    PageUp: "PageUp",
+    PageDown: "PageDown",
+    End: "End",
+    Home: "Home",
+    ArrowLeft: "ArrowLeft",
+    ArrowUp: "ArrowUp",
+    ArrowRight: "ArrowRight",
+    ArrowDown: "ArrowDown",
+    Delete: "Delete"
+  };
+
+  /*!
+   * This software or document includes material copied from or derived from https://www.w3.org/WAI/content-assets/wai-aria-practices/patterns/listbox/examples/js/listbox.js
+   * Copyright © 2023 World Wide Web Consortium. https://www.w3.org/Consortium/Legal/2023/doc-license
+   *
+   */
+  function makeAccessibleListbox(elem) {
+    elem.setAttribute("role", "listbox");
+    elem.setAttribute("tabindex", "0");
+    elem.setAttribute("aria-activedescendant", "");
+    elem.setAttribute("aria-multiselectable", "true");
+    var activeDescendantId = null;
+    var startRangeIndex = 0;
+    var keysSoFar = "";
+    var clearKeysSoFarHandle = null;
+
+    elem.addEventListener("keydown", onKeyPress);
+    elem.addEventListener("click", onClick);
+
+    return {
+      destroy: destroy,
+      clearSelectedOptions: clearSelectedOptions,
+      clearActiveOption: clearActiveOption,
+      getSelectedOptions: getSelectedOptions
+    };
+
+    function destroy() {
+      elem.removeEventListener("keydown", onKeyPress);
+      elem.removeEventListener("click", onClick);
+    }
+
+    function clearSelectedOptions() {
+      getSelectedOptions().forEach(opt => {
+        opt.classList.remove("selected");
+        opt.removeAttribute("aria-selected");
+      });
+    }
+
+    function clearActiveOption() {
+      activeDescendantId = "";
+      var activeOption = getActiveOption();
+      if (activeOption) {
+        activeOption.classList.remove("active-descendant");
+      }
+      elem.setAttribute("aria-activedescendant", "");
+    }
+
+    function getSelectedOptions() {
+      return elem.querySelectorAll('[aria-selected="true"]');
+    }
+
+    function getActiveOption() {
+      return document.getElementById(activeDescendantId);
+    }
+
+    /**
+     * Check if an item is clicked on. If so, focus on it and select it.
+     *
+     * @param {MouseEvent} evt - The click event object.
+     */
+    function onClick(evt) {
+      if (evt.target.getAttribute("role") !== "option") {
+        return;
+      }
+
+      setActiveOption(evt.target);
+      toggleSelected(evt.target);
+      scrollActiveOptionIntoView();
+
+      if (evt.shiftKey) {
+        selectRange(startRangeIndex, evt.target);
+      }
+    }
+
+    /**
+     * Handle various keyboard controls; UP/DOWN will shift focus; SPACE selects an item.
+     *
+     * @param {KeyboardEvent} evt - The keydown event object.
+     */
+    function onKeyPress(evt) {
+      var key = evt.key;
+      var lastActiveId = activeDescendantId;
+      var currentItem = getActiveOption() || getAllOptions()[0];
+      var nextItem = currentItem;
+
+      if (!currentItem) {
+        return;
+      }
+
+      switch (key) {
+        case keys.PageUp:
+        case keys.PageDown:
+          break;
+        case keys.ArrowUp:
+        case keys.ArrowDown:
+          evt.preventDefault();
+          if (!activeDescendantId) {
+            // focus first option if no option was previously focused, and perform no other actions
+            setActiveOption(currentItem);
+            break;
+          }
+
+          if (key === keys.ArrowUp) {
+            nextItem = findPreviousOption(currentItem);
+          } else {
+            nextItem = findNextOption(currentItem);
+          }
+
+          if (nextItem && event.shiftKey) {
+            selectRange(startRangeIndex, nextItem);
+          }
+
+          if (nextItem) {
+            setActiveOption(nextItem);
+            evt.preventDefault();
+          }
+
+          break;
+        case keys.Home:
+          evt.preventDefault();
+          setActiveFirstOption();
+
+          if (evt.shiftKey && evt.ctrlKey) {
+            selectRange(startRangeIndex, 0);
+          }
+          break;
+        case keys.End:
+          evt.preventDefault();
+          setActiveLastOption();
+
+          if (evt.shiftKey && evt.ctrlKey) {
+            selectRange(startRangeIndex, getAllOptions().length - 1);
+          }
+          break;
+        case keys.Shift:
+          startRangeIndex = getOptionIndex(currentItem);
+          break;
+        case keys.Space:
+          evt.preventDefault();
+          toggleSelected(nextItem);
+          break;
+        case keys.a:
+        case keys.A:
+          // handle control + A
+          if (evt.ctrlKey || evt.metaKey) {
+            evt.preventDefault();
+            selectRange(0, getAllOptions().length - 1);
+            break;
+          }
+          // fall through for Type-ahead
+        default:
+          var character = key.toUpperCase();
+          var startIndex = !keysSoFar ? getActiveOptionIndex() : null;
+
+          keysSoFar += character;
+
+          clearKeysSoFarAfterDelay();
+
+          var optionToFocus = findOption(keysSoFar, startIndex);
+          if (optionToFocus) {
+            setActiveOption(optionToFocus);
+          }
+          break;
+      }
+
+      if (activeDescendantId !== lastActiveId) {
+        scrollActiveOptionIntoView();
+      }
+    }
+
+    /**
+     * Helper function to check if a number is within a range; no side effects.
+     *
+     * @param {number} index - option index to be searched for.
+     * @param {number} start - search starts at this option index.
+     * @param {number} end - search ends at this option index.
+     * @return {boolean} - true if the number falls within the range else false.
+     */
+    function checkInRange(index, start, end) {
+      var rangeStart = start < end ? start : end;
+      var rangeEnd = start < end ? end : start;
+
+      return index >= rangeStart && index <= rangeEnd;
+    }
+
+    /**
+     * Select a range of options.
+     *
+     * @param {number} start - Options are selected from this index.
+     * @param {number} end - Options are selected till this index.
+     */
+    function selectRange(start, end) {
+      // get start/end indices
+      var allOptions = getAllOptionsArray();
+      var startIndex = typeof start === "number" ? start : allOptions.indexOf(start);
+      var endIndex = typeof end === "number" ? end : allOptions.indexOf(end);
+
+      for (var index = 0; index < allOptions.length; index++) {
+        var selected = checkInRange(index, startIndex, endIndex);
+        allOptions[index].setAttribute("aria-selected", selected + "");
+        if (selected) {
+          allOptions[index].classList.add("selected");
+        } else {
+          allOptions[index].classList.remove("selected");
+        }
+      }
+    }
+
+    /**
+     * Sets the first option as the active option.
+     */
+    function setActiveFirstOption() {
+      var firstItem = elem.querySelector('[role="option"]');
+      if (firstItem) {
+        setActiveOption(firstItem);
+      }
+    }
+
+    /**
+     * Sets the Last option as the active option.
+     */
+    function setActiveLastOption() {
+      var itemList = getAllOptions();
+      if (itemList.length) {
+        setActiveOption(itemList[itemList.length - 1]);
+      }
+    }
+
+    /**
+     * Sets the specified item as active option.
+     *
+     * @param {Element} element - The element to focus
+     */
+    function setActiveOption(element) {
+      var activeOption = getActiveOption();
+      if (activeOption) {
+        activeOption.classList.remove("active-descendant");
+      }
+      element.classList.add("active-descendant");
+      elem.setAttribute("aria-activedescendant", element.id);
+      activeDescendantId = element.id;
+    }
+
+    /**
+     * Return the previous listbox option, if it exists; otherwise, returns null.
+     */
+    function findPreviousOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = allOptions.indexOf(currentOption);
+      var previousOption = null;
+
+      if (currentOptionIndex > 0) {
+        previousOption = allOptions[currentOptionIndex - 1];
+      }
+
+      return previousOption;
+    }
+
+    /**
+     * Return the next listbox option, if it exists; otherwise, returns null.
+     */
+    function findNextOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = allOptions.indexOf(currentOption);
+
+      if (currentOptionIndex > -1 && currentOptionIndex < allOptions.length - 1) {
+        return allOptions[currentOptionIndex + 1];
+      }
+    }
+
+    function getAllOptions() {
+      return elem.querySelectorAll('[role="option"]');
+    }
+
+    function getAllOptionsArray() {
+      return Array.from(getAllOptions());
+    }
+
+    function getActiveOptionIndex() {
+      return getOptionIndex(getActiveOption());
+    }
+
+    function getOptionIndex(option) {
+      if (option == null) {
+        return -1;
+      }
+      return getAllOptionsArray().indexOf(option);
+    }
+
+    /**
+     * Toggle the aria-selected value.
+     *
+     * @param {Element} element - The element to select
+     */
+    function toggleSelected(element) {
+      var isAriaSelected = (element.getAttribute("aria-selected") === "true");
+      if (isAriaSelected) {
+        element.setAttribute("aria-selected", "false");
+        element.classList.remove("selected");
+      } else {
+        element.setAttribute("aria-selected", "true");
+        element.classList.add("selected");
+      }
+    }
+
+    /**
+     * Check if the selected option is in view, and scroll if not.
+     */
+    function scrollActiveOptionIntoView() {
+      var selectedOption = getActiveOption();
+      if (selectedOption && elem.scrollHeight > elem.clientHeight) {
+        var scrollBottom = elem.clientHeight + elem.scrollTop;
+        var elementBottom = selectedOption.offsetTop + selectedOption.offsetHeight;
+        if (elementBottom > scrollBottom) {
+          elem.scrollTop = elementBottom - elem.clientHeight;
+        } else if (selectedOption.offsetTop < elem.scrollTop) {
+          elem.scrollTop = selectedOption.offsetTop;
+        }
+      }
+    }
+
+    function findOption(searchText, startIndex) {
+      if (startIndex == null || startIndex < 0) {
+        startIndex = 0;
+      }
+
+      var allOptions = getAllOptions();
+      // Try to find after startIndex first or search behind
+      return findMatchInRange(searchText, allOptions, startIndex + 1, allOptions.length) ||
+          findMatchInRange(searchText, allOptions, 0, startIndex);
+    }
+
+    function findMatchInRange(searchText, list, startIndex, endIndex) {
+      // Find the first item starting with the keysSoFar substring, searching in
+      // the specified range of items
+      for (var n = startIndex; n < endIndex; n++) {
+        var label = list[n].innerText;
+        if (label && label.toUpperCase().indexOf(searchText) === 0) {
+          return list[n];
+        }
+      }
+      return null;
+    }
+
+    function clearKeysSoFarAfterDelay() {
+      // Restart delay if ongoing.
+      if (clearKeysSoFarHandle != null) {
+        clearTimeout(clearKeysSoFarHandle);
+      }
+
+      clearKeysSoFarHandle = setTimeout(function() {
+        keysSoFar = "";
+        clearKeysSoFarHandle = null;
+      }, 500);
+    }
+
+  }
 
   function makeAccessibleActionButton(elem, isToggleButton) {
     elem.role = "button";
@@ -104,6 +484,9 @@ define(["./_focus"], function(focusUtil) {
 
   return {
     keyCodes: keyCodes,
+    keys: keys,
+
+    makeAccessibleListbox: makeAccessibleListbox,
 
     makeAccessibleActionButton: function (elem) {
       if (elem == null) {


impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  function makeAccessibleSelect(elem) {
+    var listboxNode = elem;
```
dcleao, line 39: Why a separate variable? Does the value of either `listboxNode` or `elem` ever change?

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+     *  If there is no activeDescendant, focus on the first option
+     */
+    function setupFocus() {
+      activeDescendant = listboxNode.getAttribute(
+          'aria-activedescendant'
+      );
+      if (activeDescendant) {
+        var currentItem = document.getElementById(activeDescendant);
+        currentItem.classList.add('keyboard-navigation-focus');
+      }
```
dcleao, line 65: Here and elsewhere, it would be best to create a method such as `getActiveDescendant()` which would return the current active descendant or `null`, if none. Synchronizing the outer function scope' variable `activeDescendant`, here and elsewhere, but not always!, does not seem to provide any advantage. On the contrary, it brings confusion as to whether the variable value is expected to always be current or not.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+    var keysSoFar = '';
+    var handleFocusChange = function () {};
+    var handleItemChange = function () {};
+    registerEvents();
+
+    /**
+     * @description
+     *  If there is no activeDescendant, focus on the first option
+     */
+    function setupFocus() {
```
dcleao, line 58: This method's name `setupFocus` does not seem to match what it is doing.
You could call it `onFocus` and `onBlur`, or, to reveal what is being done, something like `setupActiveDescendantOnFocus` and `setupActiveDescendantOnBlur`.

In any case, I think a different solution would be best, perhaps. Always mark the active descendant with a specific class, e.g. `focused`. Then, in CSS, you could target that element, in the same with-focus-state, with a selector such as `:focus-visible .focused`. No additional event listening required: just update the active element's class at the same time and places when and where the `aria-activedescendant` is updated.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+
+    /**
+     * @description
+     *  Register events for the listbox interactions
+     */
+    function registerEvents() {
+      listboxNode.addEventListener('focus', setupFocus);
+      listboxNode.addEventListener('blur', setupBlur);
+      listboxNode.addEventListener('keydown', checkKeyPress);
+      listboxNode.addEventListener('click', checkClickItem);
```
dcleao, line 90: `onKeyPress`, `onClick` `onMouseDown` would be more common names.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+    var multiselectable = listboxNode.hasAttribute('aria-multiselectable');
+    var startRangeIndex = 0;
+    var keysSoFar = '';
+    var handleFocusChange = function () {};
+    var handleItemChange = function () {};
+    registerEvents();
+
+    /**
+     * @description
+     *  If there is no activeDescendant, focus on the first option
```
dcleao, line 56: JsDoc description does not appear to match what the function is doing.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+          'aria-activedescendant'
+      );
+      if (activeDescendant) {
+        var currentItem = document.getElementById(activeDescendant);
+        currentItem.classList.add('keyboard-navigation-focus');
+      }
+    }
+
+    /**
+     * @description
```
dcleao, line 69: JsDoc description does not appear to match what the function is doing.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+      listboxNode.addEventListener('keydown', checkKeyPress);
+      listboxNode.addEventListener('click', checkClickItem);
+
+      if (multiselectable) {
+        listboxNode.addEventListener('mousedown', checkMouseDown);
+      }
+    }
+
+    /**
+     * Prevent text selection on shift + click for multi-select listboxes
```
dcleao, line 98: What do you mean by "text selection"?
And, isn't shift+click allowed in multiple-selection?

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
...
+     * @param evt
+     */
+    function checkMouseDown(evt) {
+      if (
+          multiselectable &&
+          evt.shiftKey &&
+          evt.target.getAttribute('role') === 'option'
+      ) {
+        evt.preventDefault();
+      }
```
dcleao, line 109: This formatting does not follow Pentaho conventions.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,415 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  function makeAccessibleSelect(elem) {
```
dcleao, line 38: Maybe the name should be `makeAccessibleListbox`. Isn't that the ARIA/WCAG pattern being implemented? https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+        nextMatch = findMatchInRange(itemList, 0, searchIndex);
+      }
+      return nextMatch;
+    }
+
+    function findMatchInRange(
+        list,
+        startIndex,
+        endIndex
+    ) {
```
dcleao, line 365: Formatting

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+        previousOption = allOptions[currentOptionIndex - 1];
+      }
+
+      return previousOption;
+    }
+
+    /* Return the next listbox option, if it exists; otherwise, returns null */
+    function findNextOption(currentOption) {
+      var allOptions = Array.prototype.slice.call(
+          listboxNode.querySelectorAll('[role="option"]')
```
dcleao, line 395: I've seen this repeated multiple times. Sure could use a `getOptions()` method...
Also, a `getOptionIndex(option)` appears useful (like `getElementIndex` below???).

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+
+      return nextOption;
+    }
+
+    /* Return the index of the passed element within the passed array, or null if not found */
+    function getElementIndex(option, options) {
+      var allOptions = Array.prototype.slice.call(options); // convert to array
+      var optionIndex = allOptions.indexOf(option);
+
+      return typeof optionIndex === 'number' ? optionIndex : null;
```
dcleao, line 412: `optionIndex` is always a number, possibly, `-1`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+          element.classList.add('selected');
+        }
+
+      }
+    }
+
+    /**
+     * Check if the selected option is in view, and scroll if not
+     */
+    function updateScroll(key) {
```
dcleao, line 438: Only ever called with key up or down, so why not receive a boolean named, e.g. `alignTop` instead?

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+        clearTimeout(keyClear);
+        keyClear = null;
+      }
+      var keyClear = setTimeout(
+          function () {
+            keysSoFar = '';
+            keyClear = null;
+          }.bind(this),
+          500
+      );
```
dcleao, line 332: As implemented, `keyClear` is a local variable of the `clearKeysSoFarAfterDelay` fucntion.
When entering the function, its value will **always** be `undefined` and so the first `if` will never be entered.

Also, it is common to call these something with "handle" in their name, e.g. `clearKeysSoFarHandle`. This variable should be declared beside the `keysSoFar` variable to work as intended.

The `bind` call is not needed.

Formatting is off.

```js
var clearKeysSoFarHandle = null;
var keysSoFar = "";

// ...

function clearKeysSoFarAfterDelay() {
  // Restart delay if ongoing.
  if(clearKeysSoFarHandle != null) {
    clearTimeout(clearKeysSoFarHandle);
  }

  clearKeysSoFarHandle = setTimeout(function() {
    keysSoFar = "";
    clearKeysSoFarAfterDelay = null;
  }, 500);
}
```

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+
+      if (!keysSoFar) {
+        for (var i = 0; i < itemList.length; i++) {
+          if (itemList[i].getAttribute('id') == activeDescendant) {
+            searchIndex = i;
+          }
+        }
+      }
+      keysSoFar += character;
+      clearKeysSoFarAfterDelay();
```
dcleao, line 348: Could `clearKeysSoFarAfterDelay` be called from the event handler?

dcleao (changes requested):

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  function makeAccessibleSelect(elem) {
```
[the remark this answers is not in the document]
hv-vamsi: > Was this code based on someone else's?

The code is referred from standard website - https://www.w3.org/WAI/content-assets/wai-aria-practices/patterns/listbox/examples/js/listbox.js

impl/client/src/main/javascript/web/util/_a11y.js
hv-vamsi: image


After giving a though, we actually have the find feature separately. Hence removing as this is not needed.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
...
+    var startRangeIndex = 0;
+    var keysSoFar = '';
+    var handleFocusChange = function () {};
+    var handleItemChange = function () {};
+    registerEvents();
+
+    /**
+     * @description
+     *  If there is no activeDescendant, focus on the first option
+     */
```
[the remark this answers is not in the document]
hv-vamsi: image

Please share how should it look like. From the official JsDoc could not find any difference.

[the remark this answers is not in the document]
dcleao: I already provided an example (see sibling `_focus.js` file), which you could have easily used to spot the differences.

In this particular doclet (there are many other differences in other doclets in this PR), using the `@description` tag is optional and not idiomatic for Pentaho code (even for most code out there). It's seldom used. Reserved for cases where the `@classDesc` tag is also used.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,421 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  function makeAccessibleSelect(elem) {
+    var listboxNode = elem;
+    listboxNode.setAttribute('role', 'listbox');
+    listboxNode.setAttribute('tabindex', '0');
+    listboxNode.setAttribute('aria-activedescendant', '');
+    listboxNode.setAttribute('aria-multiselectable', 'true');
+    var activeDescendant = listboxNode.getAttribute(
```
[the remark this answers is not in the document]
hv-vamsi: assigned it to null

impl/client/src/main/javascript/web/util/_a11y.js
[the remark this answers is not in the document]
hv-vamsi: addressed and edited all other comments

impl/client/src/main/javascript/web/util/_a11y.js
hv-vamsi: updated with position relative on parent div and the key is removed.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+     *
+     * @param element - The element to focus
+     */
+    function setActiveOption(element) {
+      defocusItem(getActiveOption());
+      element.classList.add("activedescendant");
+      elem.setAttribute("aria-activedescendant", element.id);
+      activeDescendantId = element.id;
+
+      handleFocusChange(element);
```
dcleao, line 255: What is this function needed for? It's always a dummy function.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+        } else {
+          allOptions[index].classList.remove("selected");
+        }
+      }
+    }
+
+    /**
+     * Focus on the first option
+     */
+    function focusFirstItem() {
```
dcleao, line 225: Something like: `setActiveFirstOption`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+
+      if (firstItem) {
+        setActiveOption(firstItem);
+      }
+    }
+
+    /**
+     * Focus on the last option
+     */
+    function focusLastItem() {
```
dcleao, line 236: Something like: `setActiveLastOption`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+
+      handleFocusChange(element);
+    }
+
+    /**
+     * Defocus the specified item
+     *
+     * @param element - The element to defocus
+     */
+    function defocusItem(element) {
```
dcleao, line 263: May get rid of this function and just inline its code in `setActiveOption`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+    }
+
+    /**
+     * Focus on the specified item
+     *
+     * @param element - The element to focus
+     */
+    function setActiveOption(element) {
+      defocusItem(getActiveOption());
+      element.classList.add("activedescendant");
```
dcleao, line 251: For CSS classes, it's best to use `-` as a separator: `active-descendant`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+    /**
+     * Return the previous listbox option, if it exists; otherwise, returns null
+     */
+    function findPreviousOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = getOptionIndex(currentOption, allOptions);
+      var previousOption = null;
+
+      if (currentOptionIndex > -1 && currentOptionIndex > 0) {
```
dcleao, line 279: Isn't this the same as only `currentOptionIndex > 0`?

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+    }
+
+    function getActiveOption() {
+      return document.getElementById(activeDescendantId);
+    }
+
+    /**
+     * Check if an item is clicked on. If so, focus on it and select it.
+     *
+     * @param evt - The click event object
```
dcleao, line 71: `@param {MouseEvent} evt - ...`

Also, please, terminate statements with a `.`: `The click event object.`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+
+      if (evt.shiftKey) {
+        selectRange(startRangeIndex, evt.target);
+      }
+    }
+
+    /**
+     * Handle various keyboard controls; UP/DOWN will shift focus; SPACE selects an item.
+     *
+     * @param evt - The keydown event object
```
dcleao, line 90: `@param {KeyboardEvent} evt - ...`

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+          }
+          break;
+        case keyCodes.shift:
+          startRangeIndex = getOptionIndex(currentItem);
+          break;
+        case keyCodes.space:
+          evt.preventDefault();
+          toggleSelectItem(nextItem);
+          break;
+        case 65:
```
dcleao, line 156: Missing a key code constant? Could be called `keyCodes.A`, given it's a single letter.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      }
+    }
+
+    /**
+     * Helper function to check if a number is within a range; no side effects.
+     *
+     * @param index
+     * @param start
+     * @param end
+     * @returns {boolean}
```
dcleao, line 190: Missing parameter data types.
We use `@return` preferably to `@returns`.
Missing description text.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+
+      return index >= rangeStart && index <= rangeEnd;
+    }
+
+    /**
+     * Select a range of options
+     *
+     * @param start
+     * @param end
+     */
```
dcleao, line 204: Idem, for JsDoc

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      if (option == null) {
+        return -1;
+      }
+      return getAllOptionsArray().indexOf(option);
+    }
+
+    /**
+     * Toggle the aria-selected value
+     *
+     * @param element - The element to select
```
dcleao, line 324: `@param {Element} element - ...`
Rinse and repeat elsewhere.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      } else {
+        element.setAttribute("aria-selected", "true");
+        element.classList.add("selected");
+      }
+    }
+
+    /**
+     * Check if the selected option is in view, and scroll if not
+     */
+    function updateScroll() {
```
dcleao, line 340: Imo, a better name would be `scrollActiveOptionIntoView()`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  /*
+   * This software or document includes material copied from or derived from https://www.w3.org/WAI/content-assets/wai-aria-practices/patterns/listbox/examples/js/listbox.js
+   * Copyright © 2023 World Wide Web Consortium. https://www.w3.org/Consortium/Legal/2023/doc-license
+   *
+   */
+  function makeAccessibleListbox(elem) {
```
dcleao, line 43: Unlike the other utilities provided by this file, this one is not returning a "disposable" handle which allows the caller to cleanup the attached event listeners.

Additionally, as seen in the associated Analyzer code, some non-trivial operations are being done externally (e.g. clearing the selected options, clearing the focused option, getting the list of selected options). This could/should be provided as methods.

One way to achieve both of the above requirements would be to return an object that would expose such methods (e.g. `clearSelectedOptions()`, `clearActiveOption()`, `getSelectedOptions()`, `destroy()`).

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+    /**
+     * Handle various keyboard controls; UP/DOWN will shift focus; SPACE selects an item.
+     *
+     * @param evt - The keydown event object
+     */
+    function onKeyPress(evt) {
+      var key = evt.which || evt.keyCode;
+      var lastActiveId = activeDescendantId;
+      activeDescendantId = elem.getAttribute("aria-activedescendant");
+      var allOptions = getAllOptions();
```
dcleao, line 96: Variable not needed. Can use `getAllOptions()` instead inline in the `end` and `A` cases that use it.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+        if (selected) {
+          allOptions[index].classList.add("selected");
+        } else {
+          allOptions[index].classList.remove("selected");
+        }
+      }
+    }
+
+    /**
+     * Focus on the first option
```
dcleao, line 223: Text and punctuation: `Sets the first option as the active option.`.
Regarding punctuation, rinse and repeat.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+     */
+    function focusFirstItem() {
+      var firstItem = elem.querySelector('[role="option"]');
+
+      if (firstItem) {
```
dcleao, line 228: No new line between variable assignment statement and a following `if` testing it.
Rinse and repeat.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      element.classList.remove("activedescendant");
+    }
+
+    /**
+     * Return the previous listbox option, if it exists; otherwise, returns null
+     */
+    function findPreviousOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = getOptionIndex(currentOption, allOptions);
```
dcleao, line 276: Replace by `allOptions.indexOf(currentOption)`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      return previousOption;
+    }
+
+    /**
+     * Return the next listbox option, if it exists; otherwise, returns null
+     */
+    function findNextOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = getOptionIndex(currentOption, allOptions);
```
dcleao, line 292: Replace by `allOptions.indexOf(currentOption)`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+
+    /**
+     * Handle various keyboard controls; UP/DOWN will shift focus; SPACE selects an item.
+     *
+     * @param evt - The keydown event object
+     */
+    function onKeyPress(evt) {
+      var key = evt.which || evt.keyCode;
+      var lastActiveId = activeDescendantId;
+      activeDescendantId = elem.getAttribute("aria-activedescendant");
```
dcleao, line 95: `activeDescendantId` should only be updated by `setActiveOption(.)`.
What's the use case for doing it here/now?

dcleao (changes requested):

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
...
+      }
+      return getAllOptionsArray().indexOf(option);
+    }
+
+    /**
+     * Toggle the aria-selected value
+     *
+     * @param element - The element to select
+     */
+    function toggleSelectItem(element) {
```
dcleao, line 326: `toggleSelected`

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +35,358 @@ define(["./_focus"], function(focusUtil) {
     delete: 46
   };
 
+  /*
```
dcleao, line 38: @hv-vamsi Please, use the special comment start used for licenses: `/*!`. It is recognized and preserved by compression tools.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +36,356 @@ define(["./_focus"], function(focusUtil) {
...
+    elem.setAttribute("aria-multiselectable", "true");
+    var activeDescendantId = null;
+    var startRangeIndex = 0;
+    var keysSoFar = "";
+    var clearKeysSoFarHandle = null;
+
+    elem.addEventListener("keydown", onKeyPress);
+    elem.addEventListener("click", onClick);
+
+    return {destroy, clearSelectedOptions, clearActiveOption, getSelectedOptions};
```
dcleao, line 57: Another case of great, new JS constructs. Even though we are not unit testing this file, we should, and don't want to block that ability by using syntactical constructs that the unit test JS environment, PhantomJS used in this project does not support.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +36,356 @@ define(["./_focus"], function(focusUtil) {
...
+    /**
+     * Return the next listbox option, if it exists; otherwise, returns null.
+     */
+    function findNextOption(currentOption) {
+      // get options array
+      var allOptions = getAllOptionsArray();
+      var currentOptionIndex = allOptions.indexOf(currentOption);
+      var nextOption = null;
+
+      if (currentOptionIndex === 0 || (currentOptionIndex > -1 && currentOptionIndex < allOptions.length - 1)) {
```
dcleao, line 294: Now, `currentOptionIndex === 0` appears redundant.

hv-vamsi: addressed

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +36,386 @@ define(["./_focus"], function(focusUtil) {
...
+    return {
+      destroy: destroy,
+      clearSelectedOptions: clearSelectedOptions,
+      clearActiveOption: clearActiveOption,
+      getSelectedOptions: getSelectedOptions
+    };
+
+    function destroy() {
+      elem.removeEventListener('keydown', onKeyPress);
+      elem.removeEventListener('click', onClick);
```
dcleao, line 86: Double quotes.

impl/client/src/main/javascript/web/util/_a11y.js
dcleao: Still have not addressed this one.

hv-vamsi: addressed. removed this line

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +36,385 @@ define(["./_focus"], function(focusUtil) {
...
+      }
+    }
+
+    /**
+     * Helper function to check if a number is within a range; no side effects.
+     *
+     * @param index - option index to be searched for.
+     * @param start - search starts at this option index.
+     * @param end - search ends at this option index.
+     * @return {boolean} - true if the number falls within the range else false.
```
dcleao, line 234:
```js
@param {number} index - ...
```

Same for other parameters.

Also, for the `@return` tag, no separator `-` is/can be used.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -35,6 +36,385 @@ define(["./_focus"], function(focusUtil) {
...
+      var rangeEnd = start < end ? end : start;
+
+      return index >= rangeStart && index <= rangeEnd;
+    }
+
+    /**
+     * Select a range of options.
+     *
+     * @param start - Options are selected from this index.
+     * @param end - Options are selected till this index.
```
dcleao, line 247: `@param {number} start - ...`

dcleao (approved):
💯

hv-vamsi:
> 💯 (but, please, also fix the below minor comments I had missed!)

addressed

phalguni-a (approved):

PR #1672, opened 2023-09-06
[BACKLOG-37966] Toolbar improvements in Analyzer

AAravind27:
@pentaho/wcag  , Please review for https://hv-eng.atlassian.net/browse/BACKLOG-37966
Used in https://github.com/pentaho/pentaho-analyzer/pull/2471

### DIFF
--- a/impl/client/src/main/javascript/web/util/_a11y.js
+++ b/impl/client/src/main/javascript/web/util/_a11y.js
@@ -15,7 +15,7 @@
  * Copyright (c) 2002-2023 Hitachi Vantara. All rights reserved.
  */
 
-define(["./_focus"], function(focusUtil) {
+define("common-ui/util/_a11y", ["./_focus"], function (focusUtil) {
 
   var keyCodes = {
     A: 65,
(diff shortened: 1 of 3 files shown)


impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+    var lastItem = null;
+
+    var toolbarItems = [];
+
+    elem.addEventListener("keydown", onKeyPress);
+    elem.addEventListener("focus", onContainerFocus);
+
+    items = elem.querySelectorAll('.toolbarButton');
+
+    for (i = 0; i < items.length; i++) {
```
dcleao, line 475: `var i`

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+
+    var firstItem = null;
+    var lastItem = null;
+
+    var toolbarItems = [];
+
+    elem.addEventListener("keydown", onKeyPress);
+    elem.addEventListener("focus", onContainerFocus);
+
+    items = elem.querySelectorAll('.toolbarButton');
```
dcleao, line 473: CSS classes preferably use snake case (using `-` to separate words). Is that name taken already for other purposes?

dcleao, line 473: Why is there the need to explicitly select toolbar buttons using this class?
Would it be enough to assume that any "tabbable" descendant would be a valid toolbar element?
The only problem I can foresee is if nested tabbable elements exist.

Likewise, and if so, for determining the first, last, next or previous tabbable elements, it would be possible to directly use the `common-ui/util/_focus` API.

If it is really needed to filter tabbable elements to those with a specific selector (including both `toolbarButton` and `hidden` and `disabled`), consider extending the focus API to support that feature (e.g to receive a predicate function that determines if a given tabbable element is a valid candidate).

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+      }
+
+      if (isNotFocusable(newItem)) {
+        return getPrevious(newItem);
+      }
+      return newItem;
+    }
+
+    function setFocusToFirst() {
+      setFocusItem(firstItem);
```
dcleao, line 563: The first item should be determined dynamically, each time. Don't think the performance benefit offered by the proposed solution is worth the added complexity.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+    }
+
+    function setFocusToLast() {
+      setFocusItem(lastItem);
```
dcleao, line 567: Idem, regarding being dynamic.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+          flag = true;
+          break;
+
+        case keys.End:
+          setFocusToLast();
+          flag = true;
+          break;
+
+        case keys.Tab:
+          evt.shiftKey ? focusUtil.previousTabbable(currentItem).focus() : focusUtil.nextTabbable(currentItem).focus();
```
dcleao, line 610: Please, avoid calling side-effect functions within a conditional expression. Better to use an `if` statement, or to call `focus` on the result of the conditional.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+          flag = true;
+          break;
+
+        default:
+          break;
+      }
+
+      if (flag) {
```
dcleao, line 618: Can this variable please be named after what it controls (e.g. `handled`)?

Or, ditch it, and `return`on the `default` case.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+      var flag = false;
+      var currentItem = evt.target;
+      var elemToFocus = null;
+
+      if (!currentItem) {
+        return;
+      }
+
+      switch (key) {
+
```
dcleao, line 586: No need for this empty line.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+     *
+     * @param {KeyboardEvent} evt - The keydown event object.
+     */
+    function onKeyPress(evt) {
+      var key = evt.key;
+      var flag = false;
+      var currentItem = evt.target;
+      var elemToFocus = null;
+
+      if (!currentItem) {
```
dcleao, line 581: Better to use `== null`.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+    /**
+     * Handle various keyboard controls; UP/DOWN will shift focus; SPACE selects an item.
+     *
+     * @param {KeyboardEvent} evt - The keydown event object.
+     */
+    function onKeyPress(evt) {
+      var key = evt.key;
+      var flag = false;
+      var currentItem = evt.target;
+      var elemToFocus = null;
```
dcleao, line 579: Variable is not really needed. If you don't remove it, at least do not initialize its value, as it is being ignored in every case.

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,170 @@ define(["./_focus"], function(focusUtil) {
...
+        newItem = toolbarItems[length - 1];
+      } else {
+        newItem = toolbarItems[index - 1];
+      }
+
+      if (isNotFocusable(newItem)) {
+        return getPrevious(newItem);
+      }
+      return newItem;
+    }
```
dcleao, line 560: Not needed if focus API is used.

dcleao (changes requested):

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -458,6 +458,205 @@ define(["./_focus"], function(focusUtil) {
...
+ * Creates an accessible toolbar object for the given container element.
+ *
+ * @param {Element} elem - The toolbar container element.
+ * @param {?Object} [keyArgs] - The keyword arguments object.
+ * @param {string} [keyArgs.itemSelector] - A selector string which selects the toolbar item elements.
+ *  Defaults to `.toolbar-button`.
+ * @param {function(Element):boolean} [keyArgs.filter] - Predicate function which allows further filtering
+ *  the toolbar items. Can be used, for example, in cases of custom enabled or visible conditions.
+ * @return {({destroy: function})}} The toolbar object.
+ */
```
dcleao, line 471: Indentation is off.

dcleao (changes requested):

dcleao (approved):

hv-vamsi (approved):

PR #1677, opened 2023-09-21
[BACKLOG-37966] Toolbar improvements in PIR

AAravind27:
@pentaho/wcag Please Review


impl/client/src/main/javascript/web/dojo/pentaho/common/RowLimitControl.js
```diff
@@ -51,7 +51,10 @@ define(["dojo/_base/declare", "dijit/form/Select", "dijit/form/ValidationTextBox
             postCreate: function () {
               this.inherited(arguments);
               on(this.rowLimitRestrictions, 'change', lang.hitch(this, '_onSelect'));
+              // to support left/right navigation when this.rowLimitRestrictions is in a toolbar
```
dcleao, line 54: I think it would be better if the make accessible toolbar utility would implement this logic itself. For example, elements which have special LEFT/RIGHT arrow behaviors could specify an attribute to opt into a TAB mode, e.g. `data-pen-toolbar-nav-use-tabs="true"`.

impl/client/src/main/javascript/web/util/_focus.js
```diff
@@ -734,6 +734,21 @@ define("common-ui/util/_focus", [
...
+     * @param {Element} item - The element.
+     * @return `true` if the element is disabled; `false`, otherwise.
+     */
+    isDisabled: function (item) {
+      if ((item.ariaDisabled !== "true")) {
+        return Array.from(item.classList).some(function (className) {
+          return className.toLowerCase().includes("disabled");
+        });
+      }
+      return true;
```
dcleao, line 751: I don't think this function is of general application.

dcleao (changes requested):

impl/client/src/main/javascript/web/util/_a11y.js
```diff
@@ -644,6 +658,10 @@ define("common-ui/util/_a11y", ["./_focus"], function (focusUtil) {
       });
     }
 
+    function useTabsForNavigation(elem) {
+      return (elem && elem.dataset.penToolbarNavUseTabs) || null;
+    }
```
dcleao, line 663: Imo, this function would better return `boolean`. When the attribute is not there, should return `false`. When the attribute does not have the value `true`, should also return `false`.

impl/client/src/main/javascript/web/dojo/pentaho/common/RowLimitControl.js
```diff
@@ -284,6 +286,49 @@ define(["dojo/_base/declare", "dijit/form/Select", "dijit/form/ValidationTextBox
...
+                  });
+
+                  currentItem.setAttribute("tabindex", "-1");
+                  elem.setAttribute("tabindex", "0");
+                  elem.focus();
+                  break;
+                default:
+                  break;
+              }
+            },
```
dcleao, line 330: I'm confused. Why is this code still needed here?

dcleao (approved):
💯

varuntangirala (approved):

PR #1679, opened 2023-09-21
[BACKLOG-38666] Added Keyboard navigation for previous-next page buttons - Pentaho report viewer

kdarshana12:
@pentaho/wcag , please review for https://hv-eng.atlassian.net/browse/BACKLOG-38666
Tried to make it similar to PIR pageControl. (https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/wcag/web/src/main/javascript/pir/pir/PageControl.js)

### DIFF
--- a/impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
+++ b/impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
@@ -31,184 +31,158 @@
  *
  */
 
-define(["dojo/_base/declare", "dojo/on", "dojo/query", "dojo/_base/lang", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dijit/_WidgetsInTemplateMixin",
-        "dijit/form/NumberTextBox", "dojo/dom-class", "dojo/keys", "common-ui/util/_a11y"],
-    function (declare, on, query, lang, _WidgetBase, _TemplatedMixin, _WidgetsInTemplateMixin, NumberTextBox, domClass, keys, a11yUtil) {
-        return declare("pentaho.common.PageControl", [_WidgetBase, _TemplatedMixin, _WidgetsInTemplateMixin], {
+define(["dojo/_base/declare", "dojo/on", "dojo/query", "dojo/_base/lang", "dijit/_WidgetBase", "dijit/_TemplatedMixin", "dijit/_WidgetsInTemplateMixin", "dijit/form/NumberTextBox", "dojo/dom-class", "dojo/keys", "common-ui/util/_a11y"], function (declare, on, query, lang, _WidgetBase, _TemplatedMixin, _WidgetsInTemplateMixin, NumberTextBox, domClass, keys, a11yUtil) {
+  return declare("pentaho.common.PageControl", [_WidgetBase, _TemplatedMixin, _WidgetsInTemplateMixin], {
 
-            // Template for this widget
-            templateString: "<div class='pc_pageControlContainer' data-dojo-attach-point='containerNode'><div class='pc_pagePrev' data-dojo-attach-point='prevPageControl'></div><div class='pc_pageNext' data-dojo-attach-point='nextPageControl'></div><div data-dojo-attach-point='pageNumberInput' data-dojo-type='dijit.form.NumberTextBox' class='pc_pageNumberInput' constraints='{min:1}'></div><div class='pc_pageTotal contrast-color' data-dojo-attach-point='pageTotalSpan'>/ ##</div></div>",
+    // Template for this widget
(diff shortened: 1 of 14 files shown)


impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
```diff
@@ -30,161 +30,185 @@
...
+                }
+            },
+
+            _resetPageNumber: function () {
+                this.ignorePageNumberUpdate = true;
+                try {
+                    this.pageNumberInput.set("value", this.pageNumber);
+                } finally {
+                    this.ignorePageNumberUpdate = false;
+                }
```
dcleao, line 189: Why the try/finally block. Is there the expectation for an error to be thrown? Or, just out of extra precaution?

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
```diff
@@ -30,161 +30,185 @@
  * pc.registerPageNumberChangeCallback(controller.updatePageNumber);
  *
  */
- define(["dojo/_base/declare", "dijit/_WidgetBase","dijit/_TemplatedMixin", "dijit/_WidgetsInTemplateMixin","dijit/form/NumberTextBox", "dojo/dom-class", "dojo/keys"],
```
dcleao, line 33: Apparently, the file was completely reformatted to use 4 space tabs. While some files in pentaho/common use this rule, I'd argue they are the exception, and remains from past rules or lack thereof. Also, most other JS (and Java) codebases/areas use 2 space tabs, so I'd prefer to err on that side.

So, in this case, I would ask you to not do this. And to follow the surrounding formatting rules whenever possible.

Thanks!

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
```diff
@@ -30,161 +30,185 @@
...
+            pageNumber: 1,
+
+            // Total pages in this report
+            pageTotal: 1,
+
+            // Called once per instance
+            postCreate: function () {
+                this._resetPageNumber();
+                this.setPageCount(this.pageTotal);
+                this.own(
```
dcleao, line 58: Good move to replace the deprecated `connect` by `on` and to start using `own` to reduce memory leakage.

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
```diff
@@ -30,161 +30,185 @@
...
+                        this.changePageNumberCallback(pageNumber);
+                    }
+                    this.pageNumber = pageNumber;
+                    this.pageNumberInput.set("value", this.pageNumber);
+                    this._updatePageButtonState();
+                } finally {
+                    this.ignorePageNumberUpdate = false;
+                }
+                this.pageNumber = pageNumber;
+                this.pageNumberInput.set("value", this.pageNumber);
```
dcleao, line 165: Why not define a protected "_setPageNumberInput" method which manages the reentry control, instead of doing it everywhere the pageNumberInput is explicitly set from code?

Also, this way, the flag is set to a true value before actually needing to, making its "raison d'étre" (reason to exist) less clear.

I also find it strange that the flag is only set on the first time that the pageNumberInput is set, and not on the second time (this line). I believe it should be managed in both cases, to allow `_changePageNumber` to react only to changes initiated by the user by editing the text box directly, and to allow ignoring other code initiated changes.

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
```diff
@@ -30,161 +30,185 @@
...
+                    domClass.remove(this.prevPageControl, "pc_pagePrevDisabled");
+                    this.prevPageControl.setAttribute("tabindex", "0");
+                }
+                if (this.pageNextDisabled) {
+                    domClass.add(this.nextPageControl, "pc_pageNextDisabled");
+                    this.nextPageControl.setAttribute("tabindex", "-1");
+                } else {
+                    domClass.remove(this.nextPageControl, "pc_pageNextDisabled");
+                    this.nextPageControl.setAttribute("tabindex", "0");
+                }
```
dcleao, line 134: Regarding this, that understand to be the implementation of the roving tab index navigation pattern, in the context of being a child of toolbar, I believe that this would be better implemented by the toolbar controller itself, as noted elsewhere, where a similar implementation was proposed: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1677#discussion_r1333285474.

dcleao (changes requested):

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
dcleao: I now realize all these comments apply equally to the https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/wcag/web/src/main/javascript/pir/pir/PageControl.js you based these changes on...

impl/client/src/main/javascript/web/dojo/pentaho/common/PageControl.js
dcleao: I now realize all these comments apply equally to the https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/wcag/web/src/main/javascript/pir/pir/PageControl.js you based these changes on...

dcleao:
My comments above may apply to both `PageControl` implementations: the local one and that of PIR.

PR #1717, opened 2023-11-10
[BACKLOG-38047] WCAG - PIR -Page Zoom: General -> Template Select

singletonc:


impl/client/src/main/javascript/web/dojo/pentaho/common/TemplatePicker.js
```diff
@@ -96,7 +111,27 @@ define(["dojo/_base/declare", "dijit/_WidgetBase", "dijit/_Templated", "dojo/on"
                 a11yUtil.makeAccessibleActionButton(this.templateImg2),
                 a11yUtil.makeAccessibleActionButton(this.templateImg3),
                 a11yUtil.makeAccessibleActionButton(this.templateImg4),
-                a11yUtil.makeAccessibleActionButton(this.templateImg5)
+                a11yUtil.makeAccessibleActionButton(this.templateImg5),
+
+                //Responsive implementation
```
NJtwentyone, line 116: no change necesssary
DRY
can't you create a helper function or a loop like it does in the `showPage`

```
            showPage: function () {
              var start = this.pageNo * 6;
              for (var idx = 0; idx < 6; idx++) {

                if (idx + start < this.templates.length) {
                  this['templateImg' + idx].src = this.templates[idx + start].imagePath;
                  this['templateName' + idx].innerHTML = this.templates[idx + start].name;
                  domClass.remove(this['templateImg' + idx], 'hidden');
                  domClass.add(this['templateName' + idx], 'fade');
```

singletonc: I had assumed given that the original authors had made the loop for showPage and then manually listed each entry for the `own(...)` function, that it wasn't possible to iterate there. I know I cannot put a loop inside of that function, but let me see if I can make multiple calls to own...

singletonc: huh. I shouldn't assume. looks like I _can_ make multiple calls... I'll refactor into a loop

impl/client/src/main/javascript/web/dojo/pentaho/common/TemplatePicker.html
```diff
@@ -20,7 +20,47 @@
                         </td>
                         <td>
                             <div style="overflow:auto; width:100%; height: 100%">
-                                <table data-dojo-attach-point="templateListTbl" border="0" cellspacing="0" cellpadding="0" style="width:100%">
+                                <!--#region responsive implementation-->
```
dcleao, line 23: Ideally, the dialog would have only one set of markup. The CSS style sheets of responsive themes (ruby) would make that otherwise static markup become responsive.

Having two sets of markup, non-responsive and responsive, looks to me like a last resort solution. It has a much higher maintenance cost compared to the alternative.

I would like to better understand the reasons for adding the new set of markup. What issues were found trying to make the existing markup responsive (we have been able to make tables responsive in other instances). Alternatively, one could try to update all non-responsive supported themes to support the new markup.

Lastly, in the proposed implementation, the responsive behavior is implemented in PIR only (`pirRuby.css`), but it should be in `globalRuby.css` matching the _scope_ of this class.

dcleao (changes requested):

impl/client/src/main/javascript/web/dojo/pentaho/common/TemplatePicker.html
singletonc: As we discussed on call, I have made/tested the following changes:
- Keep the new HTML
- Get rid of the old HTML
- Cleanup js and HTML
- Move the new CSS from pirRuby.css to globalRuby.css
- Copy new CSS over to old themes
https://github.com/pentaho/pentaho-commons-gwt-modules/pull/992

dcleao (approved):

PR #1728, opened 2023-12-01
[BACKLOG-38920] Added Dot Chart

AAravind27:


impl/client/src/main/javascript/web/pentaho/visual/config.js
```diff
@@ -499,11 +499,13 @@ define(function() {
...
 
-      // Scatter
+      // Scatter, Dot
       {
         priority: RULE_PRIO_VIZ_DEFAULT,
         select: {
-          module: "pentaho/ccc/visual/Scatter"
+          module: [
+            "pentaho/ccc/visual/Scatter",
+            "pentaho/ccc/visual/Dot"]
```
dcleao, line 508: Does the same rule apply here?
A default of 5 pixels may be warranted. However, marking its value as `finished` makes it not react to/be changed due to interaction.

AAravind27: Created https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1731 that will remove the rule for Dot chart.

PR #1756, opened 2024-03-19
updating jdk17 branch with master branch changes.

amitkumar3hitachi:


AAravind27 (approved):

smmribeiro:
Created two years ago... we're already using JDK21 for our builds... I believe it is safe to consider this to be closed :)

